Issue #3, opened 2020-04-02
draw_texture_rec uses UV coordinates instead of image coordinates.


JosephCatrambone:
From drawing.rs:215:

```
let vertices = [
            Vertex::new(x    , y    , 0., sx     , sy     , color),
            Vertex::new(x + w, y    , 0., sx + sw, sy     , color),
            Vertex::new(x + w, y + h, 0., sx + sw, sy + sh, color),
            Vertex::new(x    , y + h, 0., sx     , sy + sh, color),
        ];
```

For simplicity, let us say we are always drawing to location (0,0) from a source rectangle with origin also at (0,0).  Our free parameters are 'w', 'h', 'sw', and 'sh'.  

If we are drawing a 128x128 sample at 'real' size, we set w=128, h=128, sw=128, sh=128.  Unfortunately, while the destination rectangle has the correct size (verts have 128 added for w and h), the source rectangle has UVs at (0,0), (128,0), (128, 128), (0, 128).  Instead, we should divide the UVs by the texture width and texture height.

JosephCatrambone:
No pressure on this.  If it's desirable to have the source rectangle in UV coordinates, I understand and would recommend only updating the documentation.  If it's not intended behavior, however, I made this PR: https://github.com/not-fl3/macroquad/pull/4

not-fl3:
I am totally fine with breaking changes during 0.1, thanks for pointing this out!

PR #4, opened 2020-04-02
Fix for #3.

JosephCatrambone:
Small change to make behavior of the source and destination rectangles consistent.  Source rect was initially in UV coordinates while destination rect was in screen space.  This means if you wanted your source rectangle to have a width of 128x128, you had to enter 128/texture.width(), 128/texture.height() for the source width/source height.

Another option would be to update the documentation to say that these two are different.


not-fl3:
Thanks! 

Yes, I agree, pixels for both cases are more consistent.

PR #17, opened 2020-06-14
Asteroids example

mlambir:
Hey, just created this asterodis example, also implemented the draw_poly, draw_poly_lines, draw_triangle and draw_triangle lines.
Also removed draw_hexagon since I don't see much value in it having draw_poly, but I could reimplement it using draw_poly if you think it's good to have.
One thing I'm not sure about is the signature of draw_triangle, I used 3 vec2 as parameters (similar to what raylib does) but in the rest of the draw functions we're using just f32 for x and y... maybe something to think about, I found it easier to use vec2 when i could, it has a bunch of useful methods, and adding a vec2() to the function calls if you just have f32s doesn't seem too bad for me.

Let me know what you think about it!

### DIFF
--- a/src/shapes.rs
+++ b/src/shapes.rs
@@ -107,6 +107,20 @@ pub fn draw_rectangle_lines(x: f32, y: f32, w: f32, h: f32, thickness: f32, colo
     draw_rectangle(x, y + t, t, h - t, color);
 }
 
+pub fn draw_hexagon(
+    x: f32,
+    y: f32,
+    size: f32,
+    border: f32,
+    border_color: Color,
+    fill_color: Color,
+) {
+    draw_poly(x, y, 6, size, 0., fill_color);
+    if border > 0. {
+        draw_poly_lines(x, y, 6, size, 0., border, border_color);
+    }
+}
+
 pub fn draw_poly(x: f32, y: f32, sides: u8, radius: f32, rotation: f32, color: Color) {
     let context = &mut get_context().draw_context;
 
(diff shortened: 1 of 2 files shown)


src/shapes.rs
```diff
@@ -86,98 +107,23 @@ pub fn draw_rectangle_lines(x: f32, y: f32, w: f32, h: f32, thickness: f32, colo
     draw_rectangle(x, y + t, t, h - t, color);
 }
 
-pub fn draw_hexagon(
```
not-fl3, line 89: I would like to keep `draw_hexagon` as a special case of `draw_poly`. For hex-based games :) 
And maybe later `draw_hexagon` can be extended with some additional hexagon-specific params

not-fl3:
> One thing I'm not sure about is the signature of draw_triangle, I used 3 vec2 as parameters (similar to what raylib does) but in the rest of the draw functions we're using just f32 for x and y... maybe something to think about, I found it easier to use vec2 when i could, it has a bunch of useful methods, and adding a vec2() to the function calls if you just have f32s doesn't seem too bad for me.

I think 3 `vec2` just fine. 
Right now we leak `glam` into api anyway, so lets fix it later with all the other functions with `glam` in the api. Not sure how, maybe just re-export glam from `macroquad`, but this is out of scope of this PR for sure.

mlambir:
:+1: I re-implemented draw_hexagon using draw_poly... let me know if you think that's good enough, I can bring the original back otherwise

not-fl3:
Thanks for the new awesome example! :)

not-fl3:
https://not-fl3.github.io/miniquad-samples/asteroids.html

Uploaded to miniquad-samples! :)

PR #21, opened 2020-06-23
Group layout example

Arifd:


not-fl3:
While layouts api is going to change in the future - I would prefer not to showcase it in examples right now :)

PR #31, opened 2020-07-30
DrawTextureParams docs fix

gardrek:
Docs say it's in degrees but it's in radians.


not-fl3:
Ohh, that was very confusing error in docs, thanks for the fix! :)

PR #28, opened 2020-07-22
Add config option to main macro

zotho:
- Now we can config window by send function that return [`Conf`](https://github.com/not-fl3/macroquad/pull/28/files#diff-70b83a8ee1aac5c3dc63f9734657291cR3) in `macroquad::main`

PR #32, opened 2020-08-15
Main macro optional config

mvlabat:
Based on #28 (all credit to @zotho)

Rebased against master, added a separate example for window conf.

I hope this get can merged and released soon :)


not-fl3:
Awesome, thanks!

PR #37, opened 2020-08-31
Add load_file and use it in load_texture

cedric-h:
With this, loading in game configuration is as simple as:
```rs
        let config = serde_json::from_slice(&load_file("config.json").await.unwrap()).unwrap();
```


not-fl3:
Good stuff, thank you!

PR #38, opened 2020-09-01
Make megaui optional

cedric-h:
This PR allows people to have lighter weight macroquad builds if they don't need UI, and clears the way for more PRs in the future which may involve text rendering that isn't reliant on megaui, or support for other ui frameworks in macroquad.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,44 @@ Simple and easy to use graphics library
 """
 readme="README.md"
 
+[[example]]
+name = "3d"
+required-features = [ "megaui" ] # note: only for draw_text
+
+[[example]]
+name = "asteroids"
+required-features = [ "megaui" ] # note: only for draw_text
+
+[[example]]
+name = "camera"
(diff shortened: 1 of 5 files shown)


not-fl3:
Thanks for PR, great job done here! 

However, I am not sure that I am going to merge it as is. I am going to make even more deep integration with megaui in the future - render text with the same functions, allow to use ui widgets in worls space etc, and I am not sure that I am fine with the slightly increased complexity of supporting both with megaui and without megaui versions.

cedric-h:
I understand wanting to avoid the needs of some members of the community slowing you down. My suggestion here is: don't let us slow you down. Write your code and only worry if it compiles with the `megaui` feature enabled. If members of the community, such as myself, strongly desire that macroquad work without megaui, we will come in behind you and implement that ourselves, with more PRs like this one. If no one from the community needs that, then no one writes these PRs, no harm, no foul. I'm not asking you to continue maintaining megaui-less macroquad well into the future, I'm just asking you to allow us to do that for you.

PR #39, opened 2020-09-03
Ui plugin Interface

cedric-h:


cedric-h:
After talking with Fedor, I have a better idea for how to do this, which involves adding much much less to macroquad, but potentially even provides a better interface for external UI writers and for external UI users.

PR #40, opened 2020-09-03
Add typed_characters and get_internal_mq, refactor keycode handling

cedric-h:
These are all of the changes I need to provide a nice UIPlugin system. This PR removes over twice as much code as it adds, since you said you were a huge fan of PRs with more deleted lines than added ones.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,6 +49,7 @@ struct Context {
 
     keys_down: HashSet<KeyCode>,
     keys_pressed: HashSet<KeyCode>,
+    typed_characters: String,
     mouse_pressed: HashSet<MouseButton>,
     mouse_position: Vec2,
     mouse_wheel: Vec2,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -366,6 +313,13 @@ pub fn is_key_down(key_code: KeyCode) -> bool {
     context.keys_down.contains(&key_code)
 }
 
+/// The characters typed this frame.
+///
+/// Clears after every call to next_frame().await
```
not-fl3, line 318: It looks like its not quite true - it clears after each `typed_characters` call, isnt it? 

Maybe return &[char] instead and either avoid one allocation and do get the same output on each `typed_character` during the frame?

cedric-h: Whoops, yes, I had this left over from an earlier implementation which provided all of the characters typed since the last call to `typed_characters`, which I removed because I was afraid that people would who never called `typed_characters` would end up with these huge strings allocated which they would never use.

I agree, but I think it makes more sense to return `&str` than `&[char]`.

cedric-h:
I've found a bug with this recently, because the UI renders itself directly by grabbing `get_internal_mq`, and all of the other rendering is batched, the UI will always render below everything else, which does not make for particularly useful debug ui ...

cedric-h:
I'll try implementing some sort of "hooks" system, where the `Context` saves a closure to call after rendering all of the batched stuff.

PR #51, opened 2020-10-14
Update Embark Studios logo

arirawr:
Updated to our correct logo :)


not-fl3:
Thanks! 
Borrowed that one from the Amethyst repo, maybe will make sense to ping them with the new logo as well :)

PR #53, opened 2020-10-21
Add InputFloat, rename InputField to InputString

cedric-h:


cedric-h:
https://github.com/not-fl3/megaui/pull/15 moving here

PR #57, opened 2020-10-26
Fix: _ScreenTexture dimensions not updated after window resizing

optozorax:
Without this fix:
image

With this fix:
image


not-fl3:
Thanks!

PR #59, opened 2020-11-10
dist_size no longer mandatory if source is set

puppetmaster-:
With this adjustment the setting of the **dist_size** is no longer mandatory if **source** is set.

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -124,11 +124,10 @@ pub fn draw_texture_ex(
         h: texture.height(),
     });
 
-    let (w, h) = params
-        .dest_size
-        .map_or((texture.width(), texture.height()), |dst| {
-            (dst.x(), dst.y())
-        });
+    let (w, h) = match params.dest_size{
+        Some(dst) => (dst.x(), dst.y()),
+        _ => (sw, sh),
+    };
 
     let pivot = params.pivot.unwrap_or(vec2(x + w / 2., y + h / 2.));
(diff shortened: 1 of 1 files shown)


src/texture.rs
```diff
@@ -124,11 +124,10 @@ pub fn draw_texture_ex(
         h: texture.height(),
     });
 
-    let (w, h) = params
-        .dest_size
-        .map_or((texture.width(), texture.height()), |dst| {
-            (dst.x(), dst.y())
-        });
+    let (w, h) = match params.dest_size{
```
not-fl3, line 127: You probably want to run `rustfmt` here

not-fl3:
Hmm, I do not see any behavior change with this change, am I missing something?

puppetmaster-:
Before the change this was not possible without scaling the cutout
``` rust
        draw_texture_ex(texture, player_position.x() , player_position.y(), WHITE,DrawTextureParams {
            source: Some(Rect::new(0.0,0.0,16.0,32.0)),
            ..Default::default()
        });
````

src/texture.rs
puppetmaster-: Sorry I made the adjustment directly on the github page.

not-fl3:
Ohhh, I see, `sw` is not always `texture.width()` and after the change, destination size would be the same as source texture part but not the whole texture. 

I  think this is reasonable, thanks for PR!

PR #62, opened 2020-11-13
added possibility to detect if any key was pressed

puppetmaster-:
this function allows me to write the following code
```rust
if get_last_key_pressed().is_some() {
    #[cfg(not(target_arch = "wasm32"))]
    if is_key_pressed(KeyCode::Q) | is_key_pressed(KeyCode::Escape) {
        *game_state = GameState::EXIT;
    }else{
        *game_state = GameState::GAME;
    }
}
```

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -30,7 +30,7 @@ pub fn is_key_down(key_code: KeyCode) -> bool {
 }
 
 /// Return the last pressed char.
-/// Each call "is_key_down" call will consume a character from input queue.
+/// Each "get_char_pressed" call will consume a character from the input queue.
 pub fn get_char_pressed() -> Option<char> {
     let context = get_context();
 
@@ -40,7 +40,7 @@ pub fn get_char_pressed() -> Option<char> {
 /// Return the last pressed key.
 pub fn get_last_key_pressed() -> Option<KeyCode> {
     let context = get_context();
-    //TODO adjust function to return the last key
+    // TODO: this will return a random key from keys_pressed HashMap instead of the last one, fix me later
     context.keys_pressed.iter().next().cloned()
 }
 
--- a/src/input.rs
+++ b/src/input.rs
@@ -29,12 +29,18 @@ pub fn is_key_down(key_code: KeyCode) -> bool {
     context.keys_down.contains(&key_code)
 }
 
-/// Detect if the key is being pressed.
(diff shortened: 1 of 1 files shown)


src/input.rs
```diff
@@ -29,14 +29,20 @@ pub fn is_key_down(key_code: KeyCode) -> bool {
     context.keys_down.contains(&key_code)
 }
 
-/// Detect if the key is being pressed.
-/// Each call "is_key_down" call will consume a character from input queue.
-pub fn get_key_pressed() -> Option<char> {
+/// Return the last pressed char.
```
not-fl3, line 32: I think the part of consuming characters in the docs was quite important, I would like to keep it. There was a wrong name used in the docs though, but still :) 

Also I think the `get_char_pressed` was a good enough name

src/input.rs
```diff
@@ -29,14 +29,20 @@ pub fn is_key_down(key_code: KeyCode) -> bool {
...
     let context = get_context();
 
     context.chars_pressed_queue.pop()
 }
 
+/// Return the last pressed key.
+pub fn get_last_key_pressed() -> Option<KeyCode> {
+    let context = get_context();
+
+    context.keys_pressed.iter().next().cloned()
```
not-fl3, line 43: This will return not the last pressed key but a random one from the keys_pressed hashmap. 
And to be honest, it is fine for now, but maybe add a TODO?

src/input.rs
puppetmaster-: ok like this
```rust
/// Return the last pressed char.
/// Each call "is_key_down" call will consume a character from input queue.
pub fn get_char_pressed() -> Option<char> {
.....
```

src/input.rs
puppetmaster-: is this ok?
//TODO adjust function to return the last key

src/input.rs
not-fl3: Maybe:

```
Each "get_char_pressed" call will consume a character from the input queue.
```

src/input.rs
not-fl3: I would do something like 

```
    // TODO: this will return a random key from keys_pressed HashMap instead of the last one, fix me later
    context.keys_pressed.iter().next().cloned() 
```

PR #70, opened 2020-11-26
Make a few types/mods reachable without prelude 

ozkriff:
The top-level docs look like this with these changes:

image

What do you think, @not-fl3?


ozkriff:
Update: added "color" and "math" mods.

PR #71, opened 2020-11-28
fix the filepaths for images in examples

nebkor:
Now you can run `cargo run --example=screen_texture` (for example), and it won't crash due to file-not-found.


not-fl3:
Thanks!

not-fl3:
Just FYI: paths were like this to avoid "examples" folder in the wasm build. 

But I agree that running the example on PC for the first time should be a priority!

nebkor:
Thanks for taking the pull! I hadn't considered wasm issues.

PR #89, opened 2020-12-16
Re-export exec::FileError

### LINKED ISSUE #87: FileError is returned by load_file but isn't public
There seems to be no way of matching an error in something like:

```
match load_file(path).await { ... }
```

ozkriff:
Fixes #87 (_"FileError is returned by load_file but isn't public"_)

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -2,6 +2,8 @@
 
 use crate::exec;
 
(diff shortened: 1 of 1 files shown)


ozkriff:
Thanks!

PR #93, opened 2020-12-20
chore(3d): add new draw_cube_ex render function

yamafaktory:
- Allow per side texturing (atlas / independent textures)
- Introduce new QuadTextureParams struct to pass source and rotation
- Introduce pseudo enum called Turn with variants (Quarter / Half / ThreeQuarter) to simplify rotation
- Add a new example


not-fl3:
Hi! 
I think this draw cube function is too game-specific and I am not sure that we should have it as is. 
However, I just merged "models::draw_mesh" function and with that function, you will be able to draw this kind of cube from the userspace API.

Let me know if this will help with your case!

yamafaktory:
Hi and thanks for the feedback!
I'll use this new function to adjust my code.
Let's close this PR.

yamafaktory:
@not-fl3 Just one minor thing: macroquad is still on  0.3.0-alpha.14 :bow: !

PR #104, opened 2020-12-28
added clamp to src/math.rs, polished examples/tree.rs

adsick:
Added `clamp` function in `src/math.rs`
made `example/tree.rs` look better by fixing screen ratio and adding lerped zooming with mouse wheel. 
Now it uses added `clamp`, so clamp is kinda tested.

note, that according to Rust standard libs, functions like clamp are actually made methods, actually there is f32::clamp, but it's unstable now. For better design mb we should have a trait or smth for feature like this?

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -83,3 +83,14 @@ pub fn cartesian_to_polar(cartesian: Vec2) -> Vec2 {
         cartesian.y.atan2(cartesian.x),
     )
 }
+
+/// Returns value, bounded in range [min, max].
+pub fn clamp<T: PartialOrd>(value: T, min: T, max: T) -> T {
+    if value < min {
+        min
+    } else if value > max {
+        max
+    } else {
+        value
+    }
+}


examples/tree.rs
```diff
@@ -26,19 +26,27 @@ fn tree(gl: &mut QuadGl, time: f64, deep: u32, angle: f32, tall: f32) {
...
-    };
+    let mut zoom: f32 = 0.7;
+    let mut zoominterp = zoom;
 
-    set_camera(camera);
     loop {
-        clear_background(GREEN);
-
-        draw_circle(0., 0., 0.03, RED);
+        let camera = Camera2D {
```
not-fl3, line 37: I would rather keep examples as minimal as possible, but showcasing certain features. 
Tree example is dedicated to manipulating model matrices. All the visual goodies may be done on the user side I believe :)

not-fl3: Also it is usually way easier to review PR with one certain feature in it. So it would be nice if you will split this PR into two: one just for math::clamp and another one for the example.

adsick: Nice looking exaples make good impression to Mcqd potential users. 
Hm, is it possible to fix screen ratio by somehow updating things under the hood automatically when window is being resized?
Uneven scale factors are not seeming as commonly used feature, esp for beginners, usually evertone want just a properly displayed (not stretched) image, scaled by scalar zoom factor.

not-fl3:
Thanks for PR! 
Cleaned up clamp function for a little and merged!

adsick:
Cool, this is my first contribution to github, I will make a little post in my blog about it)

PR #108, opened 2020-12-31
Cache Cargo dependencies in CI workflow

rellfy:


ozkriff:
But there's no Cargo.lock

rellfy:
> But there's no Cargo.lock

Good point. I didn't realise that. I can set Cargo.toml as a key instead, this might cause some false-positives if something other than a dependency is changed but still better than no caching I guess.

PR #135, opened 2021-01-30
Upgrade glam to 0.12

svenstaro:
Needs https://github.com/not-fl3/megaui-macroquad/pull/3. Sadly these pull requests depend on each other so it'll not compile on CI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ all-features = true
 [dependencies]
 miniquad = "0.3.0-alpha.25"
 quad-rand = "0.1"
-glam = {version = "0.10", features = ["scalar-math"] }
+glam = { version = "0.12", features = ["scalar-math"] }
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR, I definetly want to update to 0.12!

However macroquad 0.3-alpha started with glam 0.10, when .x field access was not yet implemented and the only way to get the vec field was .x()

glam 0.11 allows both ways - so updating macroquad will lead to a nice warning about .x() deprecation, wich I think is fine. 
However in 0.12 they removed .x() completely, so it is going to be a hard error. 

So I would like to postpone this PR for a little - until we will release 0.3. This way we will minimize an impact of breaking changes between glam versions.

Lets keep it open though, I will rebase this PR later on after 0.3 release :)

not-fl3:
0.3 is happening right now and `glam` finally got updated! :) 
https://github.com/not-fl3/macroquad/pull/206

svenstaro:
At last!

PR #137, opened 2021-01-31
update mouse pos even in mouse button events to fix touch pos

### LINKED ISSUE #136: touch down/up do not update mouse position
since touches are translated into mouse events by default we should also update the mouse position in `mouse_button_down_event` and co. otherwise reading the `mouse_position` after `is_mouse_button_down` is outdated. 

PR follows

extrawurst:
fixes #136


not-fl3:
Thanks!

not-fl3:
Thanks!

PR #140, opened 2021-02-01
Update image to 0.23.12

martin-t:
Advisory about previous versions: https://rustsec.org/advisories/RUSTSEC-2020-0073.html

Note that cargo audit doesn't seem to warn about it but deps.rs does which makes projects depending on macroquad look insecure.

The png_codec feature was renamed here: https://github.com/image-rs/image/commit/e706bfbb29bc4bb5c54ac644f685bcfb87d8bb07

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ all-features = true
 miniquad = "0.3.0-alpha.25"
 quad-rand = "0.1"
 glam = {version = "0.10", features = ["scalar-math"] }
-image = { version = "0.22", default-features = false, features = ["png_codec", "tga"] }
+image = { version = "0.23.12", default-features = false, features = ["png", "tga"] }
 macroquad_macro = { version = "0.1.2", path = "macroquad_macro" }
 fontdue = "0.4.0"
 bumpalo = "3.4"
--- a/src/quad_gl.rs
+++ b/src/quad_gl.rs
@@ -952,11 +952,11 @@ impl Texture2D {
         let img = if let Some(fmt) = format {
             image::load_from_memory_with_format(&bytes, fmt)
                 .unwrap_or_else(|e| panic!("{}", e))
-                .to_rgba()
+                .to_rgba8()
         } else {
             image::load_from_memory(&bytes)
                 .unwrap_or_else(|e| panic!("{}", e))
-                .to_rgba()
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -22,7 +22,7 @@ all-features = true
 miniquad = "0.3.0-alpha.25"
 quad-rand = "0.1"
 glam = {version = "0.10", features = ["scalar-math"] }
-image = { version = "0.22", default-features = false, features = ["png_codec", "tga"] }
+image = { version = "0.23.12", default-features = false, features = ["png", "tga"] }
```
not-fl3, line 25: Maybe use more relaxed "0.23" here instead?

martin-t: I am not sure. The advisory specifically mentions Patched Versions >=0.23.12. Afaik cargo selects the newest version possible when resolving deps *but* if somebody already has a version like 0.23.0 in cargo.lock, it won't update for him. Putting 0.23.12 in macroquad forces their project to update too. At least i think. OTOH the situation itself is unlikely here - somebody would have to depened on image 0.22 through macroquad and separately on 0.23 directly or through some other dep. I am wondering about this mostly for future reference not because it's likely to happen here.

Ideally, people would notice things like this because they'd have cargo audit in CI and deps.rs in readme but most projects don't. :P

I never wrote a rust lib so idk what the guidelines are - are there any? Theoretically, anything between 0.23.0 and 0.23.12 should be only compatible changes so requiring 0.23.12 shouldn't break anybody's build.

not-fl3: If we depend on 0.23.12 than someone with image 0.23.13 in their deps would have to versions of image compiled and added to the build artifacts. 
So it is really a good idea to allow cargo to select the best compatible image version :)

martin-t: I don't think so. According to https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html and https://doc.rust-lang.org/cargo/reference/resolver.html , `0.23.12` in Cargo.toml means `>=0.23.12, <0.24` so it should allow any newer version which is compatible (0.23.x where x >= 12) but not older ones with the UB.

not-fl3: ok, lets keep 0.23.12 :)

PR #152, opened 2021-02-23
material: Add ability to pass a textures to the shader

optozorax:
Before this only `Texture` and `_ScreenTexture` were accessible from the shader, but now any user texture is accessible too.


not-fl3:
Thanks, looks really good!. 

Ready to merge this when the drone will stop complaining :)

not-fl3 (approved):

not-fl3:
:tada:

PR #165, opened 2021-03-10
quad-gl: make uniforms array dynamically sized

optozorax:
* 512 is too little, it's run out of limit after couple of Mat4

Dependent PR in miniquad: https://github.com/not-fl3/miniquad/pull/183

### DIFF
--- a/src/quad_gl.rs
+++ b/src/quad_gl.rs
@@ -12,8 +12,6 @@ use std::collections::BTreeMap;
 
 //use crate::telemetry;
 
-const UNIFORMS_ARRAY_SIZE: usize = 512;
-
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, PartialEq)]
 pub struct Color {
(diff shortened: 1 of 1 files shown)


src/quad_gl.rs
```diff
@@ -738,7 +781,7 @@ impl QuadGl {
             .iter_mut()
             .zip(self.draw_calls_bindings.iter_mut())
         {
-            let pipeline = self.pipelines.get_quad_pipeline(dc.pipeline);
+            let pipeline = self.pipelines.get_quad_pipeline_mut(dc.pipeline);
```
optozorax, line 784: I am concerned about this line. Is pipeline must be immutable, and we should clone buffer on lines 829-832?

src/quad_gl.rs
```diff
@@ -783,12 +826,10 @@ impl QuadGl {
...
-            ctx.apply_uniforms(&shader::Uniforms {
-                projection: dc.projection,
-                model: dc.model,
-                time,
-                data: pipeline.uniforms_data.clone(),
-            });
+            pipeline.set_uniform("Projection", dc.projection);
+            pipeline.set_uniform("Model", dc.model);
+            pipeline.set_uniform("_Time", time);
+            ctx.apply_uniforms_buffer(&pipeline.uniforms_data[..]);
```
optozorax, line 832: Also, to use this line miniquad should be updated. I should add new version to `Cargo.toml`?

PR #166, opened 2021-03-11
input: add function `repeat_all_miniquad_input`

optozorax:
This can be useful in third-party UI libraries that run on the miniquad, like egui.

This is used in https://github.com/optozorax/egui-macroquad.

I don't quite sure that this is the right interface for a user. What do you think?


optozorax:
I think some events still missed anyway, I should re-edit this to a subscriber model, and clear `input_events` not at end of the frame, but only when a subscriber is read all his data calling function `read_all_miniquad_input`.

PR #167, opened 2021-03-12
added touch support

### LINKED ISSUE #34: Add multi-touch support on Android
Currently, you can only get the primary touch position using mouse_position().
However, most smartphones support multiple touch positions, and this feature is a must in most games.
As this feature already exists in microquad, it shouldn't be too hard to implement it in macroquad.

brianwp3000:
Resolves #34 
Added multitouch support from miniquad. I tried several different ways to structure the API, and landed on `touches() -> Vec<Touch>` because it was simple, didn't have any borrowing/lifetimes, and contained all the information needed to use touchscreen input (including the ability to track touch ids for manual gesture inputs). 

Also, the default behavior in miniquad is to treat touch events like mouse events (i.e. whenever a touch event is fired, it forwards it as a mouse event). I preserved this behavior, but added a toggle to turn it off.

I tested this by running the `input_touch.rs` example in a browser--turns out touch events aren't raised in Windows desktop apps, even if you have a touchscreen. I'm guessing that's a feature that would have to be added to miniquad.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,9 +1,9 @@
 //! Cross-platform mouse, keyboard (and gamepads soon) module.
 
-use crate::prelude::screen_width;
-use crate::prelude::screen_height;
-use crate::Vec2;
 use crate::get_context;
+use crate::prelude::screen_height;
+use crate::prelude::screen_width;
+use crate::Vec2;
 pub use miniquad::{KeyCode, MouseButton};
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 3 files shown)


brianwp3000:
Also, I didn't know if my `Touch` struct needed the NanoSerde attribute or not. Please advise

src/input.rs
```diff
@@ -75,3 +128,8 @@ pub fn is_mouse_button_released(btn: MouseButton) -> bool {
 
     context.mouse_released.contains(&btn)
 }
+
+/// Convert a position in pixels to a position in the range [-1; 1].
+fn convert_to_local(pixel_pos: Vec2) -> Vec2 {
+    Vec2::new(pixel_pos.x / screen_width(), pixel_pos.y / screen_height()) * 2.0 - Vec2::new(1.0, 1.0)
+}
```
optozorax, line 135: You could run `rustfmt` on changed files

optozorax (approved):

not-fl3:
Thanks for the PR!

I am ready to merge it, but could you run `rustfmt` on your changes? :) 

> Also, I didn't know if my Touch struct needed the NanoSerde attribute or not. Please advise
.
This is fine without NanoSerde, we generally do not use NanoSerde for macroquad's structs.

src/input.rs
brianwp3000: done

not-fl3:
Awesome, thanks again! :)

PR #185, opened 2021-04-06
fix some typos in comments

martinlindhe:
In addition to this PR, there is some more things that I think should be adjusted but it is in function and constant names and thus may cause some breakage.

The following typos was found in code, but should be addressed in a separate PR:
```
boundries => boundaries

LinearTweanFuture =>  tween

Aligment => Alignment

delimeter => delimiter

snapshoter => snapshotter   ?

```


not-fl3:
Oh, thats a lot of typos, thank you!

PR #190, opened 2021-04-10
Add missing Clone derive to AnimatedSprite

alec-deason:
It is useful to have a clone on this type so it can be used as a prototype for creating instances of an animated object.


not-fl3:
sure thing!

PR #192, opened 2021-04-11
Add documentation for Texture2D and Image.

joseluis:
These are the changes in separate commits:
- image::ImageFormat enum is exported in the prelude, since it was not really accesible before.
- Each method is now documented, plus some examples and I moved the constructors up.
- Run rustfmt on the whole library.

The examples pass unit testing. More can be added later if you like them.
I have some doubts regarding Context. The way I create it works in the examples but I don't think it's supposed to be used like that in real programs...

I think it would be good if I add documentation to some types in miniquad too.

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -23,3 +23,5 @@ pub use crate::experimental::*;
 pub use crate::logging::*;
 
 pub use crate::color_u8;
+
+pub use image::ImageFormat;
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -30,6 +30,34 @@ impl std::fmt::Debug for Image {
 }
 
 impl Image {
+    /// Creates an empty Image.
+    ///
+    /// ```
+    /// # use macroquad::prelude::*;
+    /// let image = Image::empty();
+    /// ```
+    pub fn empty() -> Image {
+        Image {
+            width: 0,
+            height: 0,
+            bytes: vec![],
+        }
+    }
+
+    /// Creates an Image from a slice of bytes that contains an encoded image.
+    ///
+    /// If `format` is None, it will make an educated guess on the
+    /// [ImageFormat][image::ImageFormat].
+    ///
+    /// # Example
+    ///
+    /// ```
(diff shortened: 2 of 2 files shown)


joseluis:
Feel free to critize anything, if you prefer to explain some things differently, please let me know.

src/quad_gl.rs
```diff
@@ -995,32 +993,46 @@ pub struct Texture2D {
     texture: miniquad::Texture,
 }
 
+/// # Constructors
```
not-fl3, line 996: We never used # Constructors / # Methods separation before and I am not really sure about that separation

joseluis: Sure, we can leave that out

joseluis:
it seems we are are pushing on top of each other :) I'm gonna leave you to edit my pull if you want, since you know better the breaking changes of today. I just deleted the comments and separation of the impls, and also dropped the rustfmt commit.

not-fl3:
> it seems we are are pushing on top of each other :) I'm gonna leave you to edit my pull if you want, since you know better the breaking changes of today. I just deleted the comments and separation of the impls, and also dropped the rustfmt commit.

I am done with textures-related changes! (hopefully)
So right now it is safe to rebase. Or I'll rebase it later myself if you want :)

joseluis:
All done. Although I'm not sure why cargo fmt fails, since I've already formatted that file... and it doesn't complain in my installation. Maybe you can take a look at that.

joseluis:
Ok, I've recreated everything over the latest mater, and now everything seems to work 100%.

I've had to update the examples in order to pass the tests, by adding the main function boilerplate on the Texture2D functions (but that part remains hidden in the docs).

joseluis:
Ok, I've recreated everything over the latest master, and now everything seems to work 100%.

I've had to update the examples in order to pass the tests, by adding the main function boilerplate on the Texture2D functions (but that part remains hidden in the docs).

not-fl3:
Thanks for PR, great work! :tada:

PR #201, opened 2021-04-14
Improve draw_rectangle_lines() so it renders well with transparency

lucypero:
`draw_rectangle_lines()` renders badly when it's transparent because it's made up of 4 overlapping rectangles. This changes that to 8 non-overlapping triangles.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,10 +1,10 @@
 on: [push, pull_request]
 
-name: Cross-compile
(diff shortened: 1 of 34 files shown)


not-fl3:
Thanks for PR, LGTM! 

But I think `is_key_released` was already merged? If so - could you rebase this from the latest master?

lucypero:
My mistake. I fixed the PR. See #204

Issue #215, opened 2021-04-28
feature-flagging rodio 


extrawurst:
since rodio was added my Wasm build blew up a few KB even though I am not making use of it and it also breaks my CI build inside the standard `rust` docker container.

PR #216, opened 2021-04-28
Audio optional

extrawurst:
closes https://github.com/not-fl3/macroquad/issues/215

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -1,23 +1,28 @@
 //! Loading and playing sounds.
 
+#[cfg(feature="audio")]
 use crate::{file::load_file, get_context};
-
+#[cfg(feature="audio")]
 use std::collections::HashMap;
 
 #[cfg(not(target_arch = "wasm32"))]
+#[cfg(feature="audio")]
 #[path = "audio/native_snd.rs"]
 mod snd;
 
 #[cfg(target_arch = "wasm32")]
+#[cfg(feature="audio")]
 #[path = "audio/web_snd.rs"]
 mod snd;
 
+#[cfg(feature="audio")]
 pub struct AudioContext {
     native_ctx: snd::AudioContext,
     sounds: HashMap<usize, snd::Sound>,
     id: usize,
 }
 
+#[cfg(feature="audio")]
 impl AudioContext {
     pub fn new() -> AudioContext {
         AudioContext {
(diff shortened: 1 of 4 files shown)


src/audio.rs
```diff
@@ -1,23 +1,28 @@
...
 
+#[cfg(feature = "audio")]
 use crate::{file::load_file, get_context};
-
+#[cfg(feature = "audio")]
 use std::collections::HashMap;
 
 #[cfg(not(target_arch = "wasm32"))]
+#[cfg(feature = "audio")]
 #[path = "audio/native_snd.rs"]
```
not-fl3, line 10: This is really a lot of #[cfg]! 
Maybe introduce new audio/dummy_snd.rs for use without the audio feature? 
It will behave a little differently - code using audio will still compiles without "audio" feature but will just skip audio output. But maybe this is fine? Especially with way simpler implementation without all the #[cfg]s?

extrawurst:
I added a dummy module `no_sound.rs` that implements our shared *interface* and reduces the `cfg`s significantly. unfortunately the `Sound::load` does not have the same signature between web and native, so there we are stuck with a few cfgs.

in general I was wondering why we do not have a `trait` for this *interface* to make sure we keep 100% the same interface to work against? but that is a different story

not-fl3:
Awesome, thanks for PR!

I like how it looks now - I believe this would be way easier to maintain the code with reduced amount of #[cfg].

Trait is a good idea, but I don't think it will help a lot - native/web implementation will still needs to be compiled with different cargo flags and therefore the only one trait impl will be used on each run.

extrawurst:
> Awesome, thanks for PR!
> 
> I like how it looks now - I believe this would be way easier to maintain the code with reduced amount of #[cfg].
> 
> Trait is a good idea, but I don't think it will help a lot - native/web implementation will still needs to be compiled with different cargo flags and therefore the only one trait impl will be used on each run.

@not-fl3 the big advantage is just that it would be checked at compile time that all implementations of the trait adhere to the same/correct *interface*

not-fl3:
I cant see how it will be checked? When compiling for web PC trait implementation will be opted out and therefore not checked, just as right now?

extrawurst:
> I cant see how it will be checked? When compiling for web PC trait implementation will be opted out and therefore not checked, just as right now?

@not-fl3 
Of course it would have to be compiled at some point with all feature permutations BUT as you can see right now: without a trait we can end up with multiple slight variations of an interface that we might want to keep consistent for all platforms. a trait would help for this and cost nothing. it communicates intent, so when I am only thinking of my Wasm target right now and add a function to `Sound` I will immediately see that I have to adjust the trait and then its easy for me to find the places that all also adhere to this trait - be it currently compiled or not. its just my habit for these things.

here is how I used that for my Macroquad web socket abstraction:
image

PR #217, opened 2021-04-29
Allow audio to be loaded from memory as well as files

alec-deason:
This breaks the `load_sound` function into two parts so that it's possible to load sound from an in memory `&[u8]` as well as a file on disk.


joseluis:
This should be synchronized with #216

alec-deason:
Yeah, totally. Should be just a matter of adding the cfg attr to the function. I'll do that as soon as the #216 merges

alec-deason:
I rebased after #216 merged but it actually looks like no changes are needed to this, so I think it's ready to go.

PR #218, opened 2021-04-30
ci: add checking particles examples

optozorax:


optozorax:
There is no way to `cd` into `particles` folder in the `uses` section to run `cargo check --examples` -_-

PR #220, opened 2021-04-30
Global error

optozorax:


joseluis:
Instead of using anyhow::Result, I'd use the following type alias:

```rust
pub type Result<T> = std::result::Result<T, Error>;
```
In order to be able to use `Result<()>` instead of  `Result<(), Error>`.

joseluis:
I think if we're gonna have a custom error type, we should probably use [thiserror](https://github.com/dtolnay/thiserror) instead of anyhow, as the creator of both crates [recommends](https://github.com/dtolnay/anyhow#user-content-comparison-to-thiserror )

optozorax:
@joseluis `anyhow` used only in one example, as `dev-dependency` to show that this can work with `anyhow` too. I can remove this and replace with `Box<dyn std::error::Error>` if `anyhow` confuses.

`thiserror` is only shorthand for writing `impl Display` and `impl std::error::Error` which is already implemented. There is no need in this dependency.

src/lib.rs
```diff
@@ -538,3 +540,53 @@ impl Window {
...
+/// ```rust ignore
+/// #[macroquad::main("Error")]
+/// async fn main() -> QuadResult {
+///     let texture = load_texture(/* ... */).await?;
+///     let material = load_material(/* ... */)?;
+///     // ...
+/// }
+/// ```
+#[derive(Debug)]
+pub enum Error {
```
not-fl3, line 559: While we use `QuadResult` notation - I think it is reasonable to call an error `QuadError`

not-fl3:
Thx!

PR #235, opened 2021-05-30
Fix profiler

MrMarble:
Currently `macroquad-profiler` is broken (not even on crates.io).

I updated the source to use `macroquad::ui` instead of `megaui-macroquad`


not-fl3:
Thanks!

PR #238, opened 2021-06-05
Add setter and clear methods to UI

henryksloan:
* Added method `set_input_focus` to `WindowContext`
* Added method `clear_input_focus` to `WindowContext`

Should setting the focus also set the window focus to the containing window, if any?


not-fl3:
Thx for PR!

PR #240, opened 2021-06-08
Re-export StyleBuilder to be visible in documentation

Bytekeeper:
Trivial change, but it allows access to the StyleBuilder in the docs.


not-fl3:
thanks!

PR #194, opened 2021-04-12
ci: add clippy

optozorax:

PR #248, opened 2021-06-24
Fix some clippy warnings and silence some with clippy.toml and config…

remilauzier:
…ure rustfmt.toml


not-fl3:
Thanks for PR! 
There is anothier PR with clippy: https://github.com/not-fl3/macroquad/pull/194

I still cant really decide what clippy.conf would be reasonable to have, it's a tough balance to catch! Cant promise that will merge either of those PRs anytime soon, but will do it eventually!

PR #258, opened 2021-07-18
tiled: Support tile flipping flags

septum:
Based from this [documentation reference](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tile-flipping).

Beware: code may be ugly and unidiomatic.

### DIFF
--- a/tiled/src/lib.rs
+++ b/tiled/src/lib.rs
@@ -1,6 +1,6 @@
-use nanoserde::DeJson;
-
 use macroquad::prelude::*;
+use nanoserde::DeJson;
+use std::f32;
 
 use std::collections::HashMap;
 
@@ -31,6 +31,13 @@ pub struct Object {
 }
 
 #[derive(Debug)]
+pub struct TileFlippingFlags {
+    pub antidiagonally: bool,
+    pub horizontally: bool,
+    pub vertically: bool,
+}
+
+#[derive(Debug)]
 pub struct Tile {
     /// id in the tileset
     pub id: u32,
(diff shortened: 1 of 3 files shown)


tiled/src/lib.rs
```diff
@@ -92,22 +101,48 @@ impl Map {
...
+                    source: Some(Rect::new(
+                        spr_rect.x - 1.0,
+                        spr_rect.y - 1.0,
+                        spr_rect.w + 2.0,
+                        spr_rect.h + 2.0,
+                    )),
+                    flip_x: flags.horizontally,
+                    flip_y: flags.vertically,
+                    rotation: if flags.antidiagonally {
+                        -90. * f32::consts::PI / 180.
```
vdrn, line 121: The issue is here. We need to multiply by 90 instead of -90.

vdrn (changes requested):
This feature requires fixing rotation code. At line 121@tiled/src/lib.rs we should multiply by `90` instead of `-90`.

vdrn:
> This feature requires fixing rotation code. At line 121@tiled/src/lib.rs we should multiply by `90` instead of `-90`.

Please disregard my previous comment.


Currrently this works for me, though I'm sure there must be  a simpler method to do this:
```
flip_x: (flags.horizontally
    && (!flags.antidiagonally || (flags.vertically && flags.antidiagonally)))
    || (!flags.horizontally && flags.antidiagonally && !flags.vertically),  
    
flip_y: flags.vertically && !flags.antidiagonally && !flags.horizontally,  

rotation: if flags.antidiagonally {  
    if !flags.horizontally {  
        -f32::consts::PI / 2.0  
    } else {  
        f32::consts::PI / 2.0  
    }  
} else {
    0.
}
```

septum:
I'll close this now.

PR #262, opened 2021-07-25
Rustaceanmark example

GeTechG:
Made it for myself, so I'm not sure it's the best solution.


darthdeus:
Just thought it would be cool to post this here, tried it and got to around ~90k sprites before FPS started to drop down from 60, running in `--release`.

image

In debug it starts dropping at around 8000. Running on AMD Ryzen 9 3900X and GTX 1080ti.

not-fl3:
Thanks, looks like a nice test!

PR #263, opened 2021-07-26
Add possibility to handle 'quit' event in main loop

aleksander:
This patch proposes a possibility for user to handle 'quit' event.
It is also required to change `miniquad::conf::Conf` struct to contain `pub handle_quit_event: bool` field.

I use this to correctly terminate networking before the quit.

### DIFF
--- a/examples/exit_dialog.rs
+++ b/examples/exit_dialog.rs
@@ -1,0 +1,43 @@
+use macroquad::prelude::*;
+use macroquad::ui::{root_ui, widgets::Window, hash};
+
+#[macroquad::main("Exit dialog")]
+async fn main() {
+
+    prevent_quit();
+
+    let mut show_exit_dialog = false;
+    let mut user_decided_to_exit = false;
+
+    loop {
+        clear_background(GRAY);
+
+        if is_quit_requested() {
+            show_exit_dialog = true;
+        }
+
+        if show_exit_dialog {
+            let dialog_size = vec2(200., 70.);
+            let screen_size = vec2(screen_width(), screen_height());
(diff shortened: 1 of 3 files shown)


examples/handle_quit.rs
```diff
@@ -0,0 +1,18 @@
...
+        handle_quit_event: true,
+        ..Default::default()
+    }
+}
+
+#[macroquad::main(window_conf)]
+async fn main() {
+    loop {
+        clear_background(WHITE);
+        if is_quit_requested() { break; }
```
not-fl3, line 15: Maybe do a `cancel_quit`(not sure about naming) API call available? To be able to do a nice  "Are you sure?" dialogue

examples/handle_quit.rs
```diff
@@ -0,0 +1,18 @@
+use macroquad::prelude::*;
+
+fn window_conf() -> Conf {
+    Conf {
+        window_title: "Handle quit event".to_owned(),
+        handle_quit_event: true,
```
not-fl3, line 6: I would avoid tweaking miniquad's Conf for this. 

I think if we will allow "prevent_quit" call from macroquad - it will be just enough to cover all the cases

If window should not close on X click:
```
if quit_requested() {
  prevent_quit();
  
  draw_exit_dialog();
}
```


If window should close on first X click, but you want to get out of the loop:
```
if quit_requested() {
  break;
}
```

And default behaviour could be the same - macroquad will just exit on close event. I think to make it happen macroquad should exit not in "exit" handler, but somewhere near "frame" callback - to let quit prevention actually works

not-fl3:
Thanks for PR! :+1:

PR #267, opened 2021-07-30
Fixes iOS path

nmpribeiro:
Small fix for iOS file loading.


nmpribeiro:
@not-fl3 , it is ready for review.

not-fl3:
thanks for PR!

Issue #244, opened 2021-06-14
`get_char_pressed` always returns `None`


vaalha:
`UiContext::process_input` consumes `get_char_pressed` as soon as the frame begins so the next caller gets None every time.

not-fl3:
Fixed with https://github.com/not-fl3/macroquad/pull/280

@GarmOfGnipahellir, thanks again :)

PR #280, opened 2021-08-15
seperate input queue for ui

GarmOfGnipahellir:
fix for #244 
this behavior might be undesirable but I needed a quick fix and decided to share

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -125,7 +125,7 @@ pub fn get_char_pressed() -> Option<char> {
     context.chars_pressed_queue.pop()
 }
 
-pub fn get_char_pressed_ui() -> Option<char> {
+pub(crate) fn get_char_pressed_ui() -> Option<char> {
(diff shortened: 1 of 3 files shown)


src/input.rs
```diff
@@ -125,6 +125,12 @@ pub fn get_char_pressed() -> Option<char> {
     context.chars_pressed_queue.pop()
 }
 
+pub fn get_char_pressed_ui() -> Option<char> {
```
not-fl3, line 128: Could you make it `pub(crate)`? 
UI will consume the entire queue anyway, so no need to make this function available in the public API

not-fl3:
Thanks for PR!

GarmOfGnipahellir:
No problem!

oofsauce:
has this been pushed to crates.io yet? i'm still hitting the bug on 0.3.8

PR #282, opened 2021-08-21
Examples: add keyboard rotation to 3d example

cjrh:
Here's what that looks like:

macroquad-3d-rotate


cjrh:
Hmm, actually this example is not that good. It would be better to use something more similar to what is in the `camera_transformations` example, but in 3d.

PR #283, opened 2021-08-23
Add jumpthough tiles to the `platformer` example

64kramsystem:
Showcase the jumpthrough functionality of the platformer crate.

The tiles have been copied from Fish Fight.

There seems to be a preexisting issue with the example physics implementation (at least on my machine) - jumping doesn't always yield the same height (so that is requires a couple of attempts to reach the wood platform), and descending the platform requires a few taps, even if the descent() function is invoked.


64kramsystem:
Closing for inactivity. Feel free to reopen if needed.

PR #287, opened 2021-09-02
Quadcraft Example

Pebaz:
* Adds a non-trivial example for 3D rendering using Macroquad with no extra dependencies.
* Cleaned up examples folder by putting assets in `res/`
* Added `draw_gizmo()` function
* Added some reusable art assets for other future examples

I wanted to add a good example for using Macroquad for 3D because there are not many examples on how to do it so I thought this would be helpful! :D


not-fl3:
Wow, very impressive work! 

I think it may deserve a separate repo with a link from a README or [awesome-quads](github.com/ozkriff/awesome-quads/)?

Pebaz:
Copy that, I'll work on getting this into it's own repo!

PR #288, opened 2021-09-02
Allow users to use arbitrary futures

### LINKED ISSUE #286: It is unsound to use regular futures within macroquad
TLDR: https://github.com/not-fl3/macroquad/blob/1f10624e820d8253b42d26ef1f1e1f616276be8e/src/exec.rs#L63 is not creating a real std::task::Context. So if the user writes e.g.

```rust
    struct Kaboom;
    impl Future for Kaboom {
        type Output = ();

        fn poll(
            self: std::pin::Pin<&mut Self>,
            cx: &mut std::task::Context<'_>,
        ) -> Poll<Self::Output> {
            cx.waker().clone(); // segmentation fault
            Poll::Ready(())
        }
    }
    CheckDone.await;
```

I believe that instead of transmute, something like the following must be used:

```rust
let data = &mut futures_context as *mut _ as *const _ as *const ();
unsafe fn clone(data: *const ()) -> RawWaker {
    RawWaker::new(data, &VTABLE)
}
unsafe fn wake(data: *const ()) {
    *(data as *const ExecState as *mut ExecState) = ExecState::Wake;
}
unsafe fn wake_by_ref(data: *const ()) {
    wake(data)
}
unsafe fn drop(data: *const ()) {
    // Nothing to do
}
const VTABLE: RawWakerVTable = RawWakerVTable::new(clone, wake, wake_by_ref, drop);
let raw_waker = RawWaker::new(data, &VTABLE);
let waker = unsafe { Waker::from_raw(raw_waker) };
let futures_context_ref = std::task::Context::from_waker(waker);
```

This will of course require modifications to the file and frame futures, too, but I'm unsure how to do that right now, will look again later this week.

Thoughts?

oli-obk:
fixes #286

In the future (haha), when we get a pending future, we could even consider sleeping/yielding the current thread until a waker has been invoked.

Some background on why I hit this:

I had to wait for a callback in some js code that I invoked. I didn't want to just loop until the value is created (not even sure that would work), so I tried implementing a future. I did a trivial one first, that didn't use wakers and just returned pending as long as there was no value. That just caused the entire macroquad state to get stuck somewhere, since I wasn't using the `ExecState` datastructure. I didn't know about that yet, so I tried implementing a waker system by sending a boxed waker to js and creating a wasm callback that js can call in order to wake the waker. So good so far, but then the wasm crashed with an index out of bounds. Some debugging later I land in the `resume` function and am trying to understand what is going on. Some research on futures, contexts and wakers later I opened #286. It wasn't quite as easy as I thought in the issue, but this PR has no user-visible changes, except that custom futures now work.

If you like this change, I could later also add some helpers (probably to sapp-jsutils) to make it easier for people to write their own futures that wrap a js callback/promise.

### DIFF
--- a/examples/back_to_the_future.rs
+++ b/examples/back_to_the_future.rs
@@ -1,0 +1,19 @@
+use std::{future::Future, task::Poll};
+
+#[macroquad::main("back to the future")]
+async fn main() {
+    struct Kaboom;
+    impl Future for Kaboom {
+        type Output = ();
+
+        fn poll(
+            self: std::pin::Pin<&mut Self>,
+            cx: &mut std::task::Context<'_>,
+        ) -> Poll<Self::Output> {
+            let cloned = cx.waker().clone(); // segmentation fault
+            drop(cloned);
+            Poll::Ready(())
+        }
+    }
+    Kaboom.await;
+}
--- a/examples/back_to_the_future_coroutines.rs
+++ b/examples/back_to_the_future_coroutines.rs
@@ -1,12 +1,44 @@
-use macroquad::prelude::{coroutines::start_coroutine, next_frame};
+use std::sync::{Arc, Mutex};
+
+use macroquad::prelude::{
+    coroutines::{start_coroutine, wait_seconds},
+    next_frame,
+};
 
 #[macroquad::main("back to the future")]
 async fn main() {
-    start_coroutine(async {
-        next_frame().await;
-        next_frame().await;
+    struct Player {
+        on_ground: bool,
+        allow_movement: bool,
+    }
+    let player = Arc::new(Mutex::new(Player {
+        on_ground: false,
+        allow_movement: false,
+    }));
+    let player2 = player.clone();
+    start_coroutine(async move {
+        loop {
+            if player.lock().unwrap().on_ground {
+                break;
+            }
+            next_frame().await;
+        }
+        println!("before wait");
+        wait_seconds(1.0).await;
+        println!("after wait");
+        player.lock().unwrap().allow_movement = true;
     });
-    for _ in 0..10 {
+    let mut i = 10;
+    loop {
+        println!("{}", i);
+        if player2.lock().unwrap().allow_movement {
+            break;
+        }
+        if i == 0 {
+            player2.lock().unwrap().on_ground = true;
+        }
+        i -= 1;
         next_frame().await;
     }
+    assert!(i < -1, "coroutine blocked main thread");
 }
--- a/examples/back_to_the_future_coroutines.rs
+++ b/examples/back_to_the_future_coroutines.rs
@@ -1,0 +1,12 @@
+use macroquad::prelude::{coroutines::start_coroutine, next_frame};
+
+#[macroquad::main("back to the future")]
+async fn main() {
+    start_coroutine(async {
+        next_frame().await;
+        next_frame().await;
+    });
+    for _ in 0..10 {
+        next_frame().await;
+    }
+}
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -1,13 +1,12 @@
 use std::future::Future;
 use std::pin::Pin;
+#[cfg(debug_assertions)]
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
 
-use crate::prelude::coroutines::{step_coroutine, Coroutine};
-use crate::MAIN_FUTURE;
-
-// "resume" sets this to true once SEEN_FRAME is true
-pub(crate) static NEXT_FRAME: AtomicBool = AtomicBool::new(false);
+// Used to detect "bad" futures being used
+#[cfg(debug_assertions)]
+static NEXT_FRAME: AtomicBool = AtomicBool::new(false);
 
 // Returns Pending as long as its inner bool is false.
 #[derive(Default)]
@@ -20,12 +19,14 @@ impl Future for FrameFuture {
     type Output = ();
 
     fn poll(mut self: Pin<&mut Self>, _context: &mut Context) -> Poll<Self::Output> {
-        if self.done && NEXT_FRAME.load(Ordering::Relaxed) {
+        if self.done {
             // We were told to step, meaning this future gets destroyed and we run
             // the main future until we call next_frame again and end up in this poll
             // function again.
             Poll::Ready(())
         } else {
+            #[cfg(debug_assertions)]
+            NEXT_FRAME.store(true, Ordering::Relaxed);
             self.done = true;
             Poll::Pending
         }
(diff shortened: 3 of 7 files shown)


not-fl3:
Thank you very much for this PR! 

It was a really bad implementation and I am glad it is finally fixed.

Actually I even had the exactly same problem with JS once, but I just postponed the fix for later, and this later never came :)

not-fl3:
I found a problem with this approach. 

Before, each future had its own, individual context: 
```rust
pub(crate) struct CoroutinesContext {
    futures: Vec<Option<(Pin<Box<dyn Future<Output = ()>>>, ExecState)>>,
}
```

And it was possible to "tick" any future at any given moment, they were not really connected to frames. 

But now we got a global state, shared by all the futures. 

I found it out from `"invalid state in frame static"` in a game that heavily utilizes macroquad's coroutines, will do a publicly available repro later.

oli-obk:
Ah yes, I was worried about this when I touched coroutines, but could not cause any problems with the examples.

oli-obk:
minimal repro:

```
use macroquad::prelude::{coroutines::start_coroutine, next_frame};

#[macroquad::main("back to the future")]
async fn main() {
    start_coroutine(async { next_frame().await; next_frame().await; });
    for _ in 0..10 {
        next_frame().await;
    }
}
```

using `next_frame()` within a coroutine as the second await site will cause this

oli-obk:
What is the expected behaviour here? I thought next_frame was just to "wait" until the next frame, which coroutines do for any future that returns Pending. Can we just outlaw next_frame in coroutines?

oli-obk:
Ok :D I figured it out. Coroutines work exactly like the main loop now, they even use `resume`.

not-fl3:
> What is the expected behaviour here? I thought next_frame was just to "wait" until the next frame, which coroutines do for any future that returns Pending. Can we just outlaw next_frame in coroutines?

Coroutines have a few various methods to postpone the execution, `next_frame` stops execution until next frame, `wait_for_seconds` stops execution for exact time. 

`next_frame()` usually is used to wait for something to happen, like this:

```rust
while !player.on_ground() {
  next_frame().await;
}
play_fall_sfx();
wait_for_seconds(0.1).await;
player.allow_movement = true;
```

Similar coroutines are often used to encode sequences of visual effects and logic, like spawn a bullet, wait until it reaches the target, play sfx and particle effect, wait for both, and, maybe, reduce the target's healthpoints.

There is a missing and not yet implemented feature, but this feature may affect coroutines design: coroutines should be "pausable", they should have their own, individual timelines. 

As an example: 
`start_coroutine(async { wait_for_seconds(2.).await; })` right now will be "ticked" inside macroquad and will always wait for 2 seconds. 
But imagine if the game is paused while coroutine is still active. 
If this coroutine is a GUI coroutine responsible for the pause menu itself(idk, scaling buttons on mouse hover) - it should keep going. But if it is part of a game logic - it should pause and wait for the unpause. Right now there is no way to implement this, each coroutine can be either stopped completely or will be executed till the end.

So, I believe, there should be a way to attach some sort of a context to each of rust's futures, and we can't really rely only on a global shared state for controlling the execution.

oli-obk:
So... I have fixed the issue with coroutines blocking until the next frame. It does not support sub-frame pending operations. All pending operations are delayed until the next frame.

> But if it is part of a game logic - it should pause and wait for the unpause. Right now there is no way to implement this, each coroutine can be either stopped completely or will be executed till the end.

this is a twofold problem:

1. the wait_seconds future takes the global time via `miniquad::date::now()`, so it won't care about pausing anyway.
    * this can be resolved by pre-computing the number of frames that this timer should take and counting the frames within the future.
2. Pausing the "stepping" of a future.
    * this can be resolved by adding a bool field (`paused`) to the value inside `CoroutinesContext::futures`. We can then use that field to decide whether to step the future at all.

oli-obk:
> So, I believe, there should be a way to attach some sort of a context to each of rust's futures, and we can't really rely only on a global shared state for controlling the execution.

this global state is now solely used for figuring out when to pause the main thread. I don't see a good way around that, but since we reset the global state before stepping the main thread, I don't see any problems with this.

not-fl3:
Just tried it on the fish game (which used coroutines a lot) and it works perfectly well! 

But I got a new question :)

We cant block on WASM - browser will not fire any callbacks until we will exit handling our `animationFrame`. 
And it looks like the new `resume` function will block main thread waiting for `NEXT_FRAME`? So on WASM we will block forever :(

And the second problem: 
```
load_file(a).await;
load_file(b).await;
```

if files a and b are real small - waiting for an extra frame on each `onload` may take some noticeable time.

not-fl3:
Oh, the second problem is not a problem at all, it actually will be even faster than before, edited comment :)

oli-obk:
> We cant block on WASM - browser will not fire any callbacks until we will exit handling our `animationFrame`.
> And it looks like the new `resume` function will block main thread waiting for `NEXT_FRAME`? So on WASM we will block forever :(

how did that work before?  `resume` is only ever called once per frame, so we can't use it to bounce back to wasm and then re-invoke resume until we hit a `next_frame`, right? I can easily change it to just return from `resume` on every pending operation, but that means every pending operation will "block" until the next frame.

oli-obk:
I somewhat fixed it ("works for me"). We will now yield to js on every single pending future. But we only do that on wasm, all other platforms will still happily munch away on the futures until a frame future is encountered. I have ideas how to do the same for wasm, will experiment now.

oli-obk:
Ok... 9b6f167 introduces a system that will fall on our feet if someone uses threads to wake futures, but works totally fine on wasm.

The main problem is that the global context is not thread safe as far as I can tell, so that using threads to wake futures will do things that are not thread safe. We can fix this by adding locking to MAIN_FUTURE and step_coroutine.

Considering that it was impossible to use wakers before this PR, no one is using threads (or anything else) to wake futures.

So, if the above design works for you (it does for me, but I don't use coroutines yet), then I can add the appropriate locking mechanisms on non-wasm.

not-fl3:
I just want to say that I love this PR, learned so much :heart: 

Took me some time to grasp all the new knowledge :)
I tried it with all the games I have access to - and it works just fine. 

> how did that work before?

I see that you already figured this :) It worked exactly like in your last change - `resume` was always polling future exactly once, without blocking or waiting. If not ready - just try again on the next frame. 

About global state in weaker - I am really not so sure about this
Maybe leave `Waker::wake` empty and explicitly not allow macroquad's main future being waked by anything but `exec::resume`?  This will simplify implementation a lot and will give lots of freedom for coroutines implementation

Main future should be polled on an exact moment within a frame, waking it at any other moment will break things anyway, so, maybe, we should allow waiting for user futures within macroquad's main loop, but not bother with allowing waking the future itself from the outside?

oli-obk:
> Took me some time to grasp all the new knowledge :)

I will add more documentation

> Main future should be polled on an exact moment within a frame, waking it at any other moment will break things anyway

ah, that is a problem indeed. Now I understand. The same thing applies to coroutines, as they can also draw and such.

using std::thread will also go horribly wrong if I'm not wrong?

I'll fix all of this, that makes everything simpler indeed

not-fl3:
Going to merge this, thanks for your work! Macroquad needed this change :)

not-fl3:
The only thing I would maybe do later - move futures tests from /examples to /tests

But `#[macroquad(main)]` macro will require some changes to make it happen and it is definitely outside of the scope of this PR :)

not-fl3:
> The only thing I would maybe do later - move futures tests from /examples to /tests
> 
> But `#[macroquad(main)]` macro will require some changes to make it happen and it is definitely outside of the scope of this PR :)

Done, now futures tests lives in /tests: https://github.com/not-fl3/macroquad/tree/master/tests :)

oli-obk:
sweet! that's really cool

PR #289, opened 2021-09-03
Remove broken assertions and unused impls

oli-obk:
you merged too fast :D The debug assertions don't actually work. We'd have to manually support them in each future and that is a pain. Let's just remove them


not-fl3:
Oh, my bad, thanks for PR!

PR #290, opened 2021-09-07
coroutines: Support pausable coroutines

not-fl3:
Syntetic tests: [tests/coroutine_pause.rs](https://github.com/not-fl3/macroquad/blob/coroutine_pause/tests/coroutine_pause.rs)

Less syntethic test: 

https://user-images.githubusercontent.com/910977/132287061-f7f9c54b-fd4c-4a73-8e40-54b7d2057b6b.mp4

This change allows switching coroutines into a "manual" polling mode. 
Each coroutine in "manual" mode has its own timeline and could be advanced with given delta time - so they are not dependent on real frames deltas. It may be helpful for simulation pause, "fixed" delta time simulation etc.

@oli-obk The thing with pausable coroutines I was talking about in your futures PR.  It works! 

Ended up with some `get_context` calls inside `wait_seconds`, but still better than passing context through unsound transmuting

### DIFF
--- a/src/experimental/coroutines.rs
+++ b/src/experimental/coroutines.rs
@@ -10,21 +10,31 @@ use std::task::{Context, Poll};
 use crate::exec::resume;
 use crate::get_context;
 
+struct CoroutineInternal {
+    future: Pin<Box<dyn Future<Output = ()>>>,
+    manual_poll: bool,
+    manual_time: Option<f64>,
+}
+
 pub(crate) struct CoroutinesContext {
-    futures: Vec<Option<Pin<Box<dyn Future<Output = ()>>>>>,
+    coroutines: Vec<Option<CoroutineInternal>>,
+    active_coroutine_now: Option<f64>,
+    active_coroutine_delta: Option<f64>,
 }
 
 impl CoroutinesContext {
     pub fn new() -> CoroutinesContext {
         CoroutinesContext {
-            futures: Vec::with_capacity(1000),
+            coroutines: Vec::with_capacity(1000),
+            active_coroutine_now: None,
+            active_coroutine_delta: None,
         }
     }
 
     pub fn update(&mut self) {
-        for future in &mut self.futures {
+        for future in &mut self.coroutines {
             if let Some(f) = future {
-                if resume(f) {
+                if f.manual_poll == false && resume(&mut f.future) {
                     *future = None;
                 }
             }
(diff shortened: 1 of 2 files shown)


oli-obk:
An alternative design would be to create a "pausable" wrapper like

```rust
async fn pausable(paused: Arc<AtomicBool>, future: impl Future<Output = ()> + Send + 'static) {
    struct Pausable<T: Future<Output = ()> + Send + 'static>(Arc<AtomicBool>, inner: T);
    impl<T: Future<Output = ()> + Send + 'static> Future for Pausable<T> {
        type Output = ();
        fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
            if !self.0.load(Ordering::Relaxed) {
                self.1.poll(cx)
            } else {
                Poll::Pending
            }
        }
    }
    Pausable(future) 
}
```

Then there can be multiple pausing deciders, and they can be shared between coroutines. This has the further advantage that you don't need to touch each coroutine when you pause, but you only touch them once as "pausable" at creation time. It also keeps the coroutine handling simple, as all the complexity is in the wrapper.

src/experimental/coroutines.rs
```diff
@@ -91,8 +168,13 @@ impl Future for TimerDelayFuture {
...
 
 pub fn wait_seconds(time: f32) -> TimerDelayFuture {
+    let now = get_context()
+        .coroutines_context
+        .active_coroutine_now
+        .unwrap_or_else(|| miniquad::date::now());
+
     TimerDelayFuture {
-        start_time: miniquad::date::now(),
+        start_time: now,
```
oli-obk, line 177: Alternatively you could remove `start_time` and replace it with `remaining_time`. Then, in each tick, you do `self.remaining_time -= get_frame_time()` and check if it is negative to decide whether to finish the future or keep pending

not-fl3: Thanks, fixed, `remaining_time` is better!

not-fl3:
> An alternative design would be to create a "pausable" wrapper like
> 
> ```rust
> async fn pausable(paused: Arc<AtomicBool>, future: impl Future<Output = ()> + Send + 'static) {
>     struct Pausable<T: Future<Output = ()> + Send + 'static>(Arc<AtomicBool>, inner: T);
>     impl<T: Future<Output = ()> + Send + 'static> Future for Pausable<T> {
>         type Output = ();
>         fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
>             if !self.0.load(Ordering::Relaxed) {
>                 self.1.poll(cx)
>             } else {
>                 Poll::Pending
>             }
>         }
>     }
>     Pausable(future) 
> }
> ```
> 
> Then there can be multiple pausing deciders, and they can be shared between coroutines. This has the further advantage that you don't need to touch each coroutine when you pause, but you only touch them once as "pausable" at creation time. It also keeps the coroutine handling simple, as all the complexity is in the wrapper.

Saved the snippet, but, for this case, I like the `manual_poll` a little more - it is easier to control the exact poll moment for coroutines, and it helps with my simulation pause case.

oli-obk:
>  it is easier to control the exact poll moment for coroutines, and it helps with my simulation pause case.

oh, good point, I didn't think of that.

PR #297, opened 2021-09-23
Improve docs and fix Into<Camera2D> for `experimental::camera::mouse`

SpyrosRoum:
- I changed the variable name for `mouse_pos` to `center` because I think using `mouse_pos` suggests that the mouse position should be used, even thought that's not necessary
- I changed `impl Into<Camera2D> for Camera` to `impl Into<Camera2D> for &Camera`, so it doesn't have to be cloned on each frame
- I derived `Copy` for Camera because it's trivial to do
- I improved the documentation with some suggestions and examples


not-fl3:
Thanks!

Issue #30, opened 2020-07-29
Blurry text using `draw_text`


Catman155:
When using the `draw_text` method, the text drawn is blurry at higher font sizes. Doe this have to do with the fact that a picture is used as a font? Is there a way to get a better looking text at higher font sizes?

luciusmagn:
Hi, I'm experiencing this as well, and I think it concerns pretty much all sizes

gubbu:
Its the filter Mode:
pub enum FilterMode {
    Linear = GL_LINEAR as isize,
    Nearest = GL_NEAREST as isize,
}
When the texture mode is set to "Linear" (the default) you get blurred results of the pixel fonts... I guess the problem is: context.font_texture FilterMode is set to "Linear" instead of Nearest?

extrawurst:
> Its the filter Mode:
> pub enum FilterMode {
> Linear = GL_LINEAR as isize,
> Nearest = GL_NEAREST as isize,
> }
> When the texture mode is set to "Linear" (the default) you get blurred results of the pixel fonts... I guess the problem is: context.font_texture FilterMode is set to "Linear" instead of Nearest?

Nearest will only make it more pixelated and jaggy. in the end the font texture resolution would need to be bigger to prevent this

SrMordred:
Any updates regarding this ? 
If not, any points on how to solve this ?
I can help with PR (although i'm not experienced )

PR #310, opened 2021-10-24
Add high-dpi support

dotellie:
Fixes https://github.com/not-fl3/macroquad/issues/30

This PR worked in my very contrived scenario so what I would encourage everyone who had the problem in #30 to do is test this out locally and report errors here! I have very likely missed places where the raw screen size is used without conversion. Add this to your `Cargo.toml` to test: `macroquad = { git = "https://github.com/dotellie/macroquad.git", branch = "high-dpi" }`

## Comparison

Before
Screenshot 2021-10-24 at 17 22 54

After
Screenshot 2021-10-24 at 17 22 11

Funny thing is actually that the circle was also affected by this problem as you can see in those screenshots 😅

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -681,7 +681,7 @@ impl Window {
             conf::Conf {
                 sample_count: 4,
                 window_title: label.to_string(),
-                // high_dpi: true,
+                high_dpi: true,
                 ..Default::default()
             },
             future,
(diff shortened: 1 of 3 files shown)


mlindner:
Please accept this change (assuming it's done). This is rather annoying for anyone on Mac OS that makes this library rather annoying to use.

dotellie:
Hey @mlindner! I would basically consider this done, but I am waiting for both people like you to confirm that this change doesn't break your game/app and for @not-fl3 to go through the changes and give any feedback they may have. After all, they know how macroquad rendering works and I... don't really 😅

not-fl3:
I just never had any high-dpi devices so my understanding of what really is high-dpi is quite limited.

But apparently, this really helps people, so lets get it merged!

PR #316, opened 2021-11-13
Add docs to shapes & rectangle

darthdeus:
Happy to make edits to anything, I just tried to document the stuff that was initially unclear to me that I had to try out, such as if `Rect` is defined by its center or top-left.


not-fl3:
Awesome, thanks!

PR #337, opened 2021-12-12
Add function to display tiny fonts without anti-aliasing.

Roxfall:


Roxfall:
Premature pull request, not working right, testing.

PR #343, opened 2021-12-18
Scale mouse position with dpi

### LINKED ISSUE #331: button hitbox is offset from where the button actually is
using this code 
```rust
use macroquad::prelude::*;
use macroquad::math::vec2;
use macroquad::ui::{root_ui, widgets};
#[macroquad::main("button testing")]
async fn main() {
    loop {
        clear_background(BLACK);
        if widgets::Button::new("push me").size(vec2(100.0, 100.0)).position(vec2(100.0, 100.0)).ui(&mut *root_ui()) {
            println!("button was pushed");
        }
        next_frame().await;
    }
}
```
This shows up
The yellow rectangle was added in by me to show approximately where the button could be clicked, and it doesn't line up with the actual button
button testing
The hitbox (clickbox?) seems to have different offsets if I move the button to different parts of the screen. However, it always offsets towards the top left of the button.
any help?

vpzomtrrfrt:
Should fix #331

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -50,7 +50,10 @@ pub fn show_mouse(shown: bool) {
 pub fn mouse_position() -> (f32, f32) {
     let context = get_context();
 
-    (context.mouse_position.x, context.mouse_position.y)
+    (
+        context.mouse_position.x / context.quad_context.dpi_scale(),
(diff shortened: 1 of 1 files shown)


not-fl3:
Oh, nice, good catch, thank you!

PR #345, opened 2021-12-21
Fix: incorrect Circle::contains check

pdcgomes:
The point in circle calculation was checking if the incoming point was *outside* the circle instead of inside.


not-fl3:
Thanks!

PR #350, opened 2021-12-25
Deleting old atlas texture when creating a new one

### LINKED ISSUE #320: Atlas texture should be deleted when creating new one
https://github.com/not-fl3/macroquad/blob/8e80940d611e09cbb360a52c2550a8fed866e9e5/src/text/atlas.rs#L81 creates a new atlas texture. If `self.texture` is already set, it should be `.delete()`d.

MarcelCoding:
Closes #320


not-fl3:
Thanks!

PR #355, opened 2021-12-27
Implemented shadowed text

### LINKED ISSUE #354: Add text shadow

MarcelCoding:
Closes #354 

I updatet the text example:
image


MarcelCoding:
@not-fl3 
Could you maybe try a rerun for the one that had a timeout?

not-fl3:
I like it! 

But I am not exactly sure if this will work well with all the fonts and all the cases. As for me this solution (draw a text twice to emulate the shadow) - is a perfectly fine technique for a userspace code, but I am not really sure if this belongs to the engine. 

I do not really have a strong opinion on this, so I cant really merge or decline this

MarcelCoding:
I thought about this because you either had to draw the text every tome two times or do a wrapper method. Another point was that there is some calculation being done, the calculated values could be shared with the shadow. (This isn't really a reason - the impact should be very very low.)

PR #359, opened 2022-01-01
added `mouse_wheel_normalized()`

### LINKED ISSUE #347: `mouse_wheel()` is inconsistent between targets
On Linux, `mouse_wheel()`'s elements are -1.0, 0.0, or 1.0, but on my machine in Firefox through WASM, they are -114.0 and 114.0.

It might be beneficial to keep this since I assume that corresponds to something like my Firefox scroll speed, but it should be documented and maybe a normalized function should be provided.

mkhan45:
Fixes #347 

Added a `mouse_wheel_normalized()` function:
```rs
/// Returns the x and y mouse scroll normalized to -1.0, 0.0, or 1.0
pub fn mouse_wheel_normalized() -> (f32, f32) {
    let context = get_context();

    let normalize = |v: f32| if v == 0.0 { 0.0 } else { v.signum() };

    (
        normalize(context.mouse_wheel.x),
        normalize(context.mouse_wheel.y),
    )
}
```

I'm not sure if return an enum would be better like so:
```rs
enum MouseScroll {
    ScrollUp,
    ScrollDown,
    NoScroll,
}
```

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -93,10 +93,23 @@ pub fn touches_local() -> Vec<Touch> {
         .collect()
 }
 
+/// Returns the x and y mouse scroll, unnormalized
 pub fn mouse_wheel() -> (f32, f32) {
     let context = get_context();
 
     (context.mouse_wheel.x, context.mouse_wheel.y)
+}
+
+/// Returns the x and y mouse scroll normalized to -1.0, 0.0, or 1.0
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR! 

I believe the proper fix for #347 should be in miniquad, equalizing mouse scroll speed over platforms.
Using signum as a workaround is totally fine, but I am not really sure if we should add a separate function with a signum to the macroquad itself. `

PR #363, opened 2022-01-05
Added window::request_new_screen_size

VictorKoenders:
Exposed miniquad's `set_window_size` as a function. I'm using this to test that my game runs on different phone sizes correctly without having to rebuild my game every time.

While testing this on Windows I immediately noticed that Windows doesn't always respect the value that you feed it. In my case I fed it a value of 3040x1440 (the screen size of a Galaxy Note 10+) and windows only gave me 3040x1061. This is probably dependent on the monitor layout. 

Because of this inconsistency (and presumably many more OS-specific quirks) I decided on a very conservative name and a explanation on the behavior. I also didn't want to update `context.screen_width/height` immediately because then the value would be `A -> Requested B -> frame end -> Actual B`, which seemed more confusing to me.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -65,6 +65,20 @@ pub fn screen_height() -> f32 {
     context.screen_height / context.quad_context.dpi_scale()
 }
 
+/// Request the window size to be the given value. This takes DPI into account.
+///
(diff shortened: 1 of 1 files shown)


not-fl3:
Nice, thanks for PR! :+1:

Issue #333, opened 2021-12-09
`fn get_context() -> &'static mut` is unsound


parasyte:
This issue was brought to my attention in https://www.reddit.com/r/rust/comments/rcb8wg/are_static_mut_really_that_bad/ where a commenter (now deleted by the author) used Macroquad's usage of `static mut` as an example of how it can be used without worry.

https://github.com/not-fl3/macroquad/blob/33a8e4a3eca7c323ab09042c35a6b056bb74b773/src/lib.rs#L414-L416

After looking over the code for a few minutes, it became apparent that the use of `static mut` here is most certainly unsound, in _at least_ this case: https://github.com/not-fl3/macroquad/blob/33a8e4a3eca7c323ab09042c35a6b056bb74b773/src/lib.rs#L503-L533

The issue is that `touch_event` acquires a unique borrow of globally accessible state, holding the reference across a function call boundary that itself also acquires the same unique borrow. I haven't been able to write a test that calls the `touch_event` method, so I cannot provide proof of unsoundness with a test. Instead, I wrote a minimal test case based on the structure of code that illustrates the problem: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c94caef0ca7aa1aa57c26ed81326a521

Miri output for the minimized test case:

```
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.87s
     Running `/playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo-miri target/miri/x86_64-unknown-linux-gnu/debug/playground`
error: Undefined Behavior: no item granting read access to tag <2025> at alloc1+0x8 found in borrow stack.
  --> src/main.rs:27:5
   |
27 |     context.0 += 1;
   |     ^^^^^^^^^^^^^^ no item granting read access to tag <2025> at alloc1+0x8 found in borrow stack.
   |
   = help: this indicates a potential bug in the program: it performed an invalid operation, but the rules it violated are still experimental
   = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
           
   = note: inside `touch_event` at src/main.rs:27:5
note: inside `main` at src/main.rs:34:5
  --> src/main.rs:34:5
   |
34 |     touch_event();
   |     ^^^^^^^^^^^^^
[...]

PR #365, opened 2022-01-09
[WIP] Use RacyCell for global CONTEXT

tmfink:
Starts to address #333.

Mutable static variables are very difficult to use soundly. Uses a `RacyCell` as suggested in
https://github.com/rust-lang/rust/issues/53639#issuecomment-790091647 with some improvements from https://docs.rs/cortex-m-rtic/1.0.0/src/rtic/lib.rs.html#89.

Using a `RacyCell` requires some "whole-program" analysis to ensure that we don't have aliasing mutable pointers, but it's more tractable than using mutable statics directly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ bumpalo = "3.4"
 backtrace = { version = "0.3.60", optional = true, default-features = false, features = [ "std", "libbacktrace" ] }
 log = { version = "0.4", optional = true }
 quad-snd = { version = "0.2", optional = true }
+static_assertions = "1.1"
 
 [dev-dependencies]
 macroquad-particles = { path = "./particles" }
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -96,7 +96,9 @@ pub async fn load_sound(path: &str) -> Result<Sound, crate::file::FileError> {
 /// Attempts to automatically detect the format of the source of data.
 pub async fn load_sound_from_bytes(data: &[u8]) -> Result<Sound, crate::file::FileError> {
     let sound = {
-        let ctx = &mut get_context().audio_context;
+        // SAFETY: using raw pointer to hand out mut ref as `ctx`, which avoids aliasing
+        let ctx = unsafe { &mut (*get_context()).audio_context };
+
         QuadSndSound::load(&mut ctx.native_ctx, data)
     };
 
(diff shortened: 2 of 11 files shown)


tmfink:
Can we use a sound abstraction that does not require manual "whole-program" analysis like I have now?

If we want a global variable approach, then we need
- Something to provide `Sync`
- Something to provide mutability or interior mutability

For example, a [`parking_lot::Mutex`](https://docs.rs/parking_lot/0.11.2/parking_lot/type.Mutex.html):

```rust
static CONTEXT: parking_lot::Mutex<Option<Context>> = parking_lot::const_mutex(None);
```

Also, do we need to use a global variable?
Could we create the `Context` in the "real" `main()` function and pass it to the user in their "fake" `main()` function? This would avoid the performance (and syntactical overhead) of something like a `Mutex` or `RacyCell`.

tmfink:
I'm closing this work-in-progress PR since I think the approach with `RacyCell` does not scale. There is too much unsafe code that would be too difficult to write/maintain. Safe methods should be used instead.

I'll open a separate PR.

# Other methods that don't work

## `parking_lot::Mutex`
The naive `parking_lot::Mutex` stored in a global does not work well since it's easy to get deadlocks. For example, if a user writes a helper function that calls `get_context()` which locks the Mutex when the Mutex is already locked in the parent function frame, then you get a deadlock (even though there is only one thread).

[Mutex with deadlock example playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0cabe788afb2d065ea4a9256748514b8)

## `parking_lot::ReentrantMutex`

In order to avoid the deadlock scenario above, we could try to use a [`parking_lot::ReentrantMutex`](https://docs.rs/parking_lot/0.11.2/parking_lot/type.ReentrantMutex.html). While I was able to get this to partially work, the major issue is that the `ReentrantMutex` does not provide interior mutability. According to the docs:

> `ReentrantMutexGuard` does not give mutable references to the locked data. Use a `RefCell` if you need this.

Hence, the declaration for `CONTEXT` would be:

```rust
static CONTEXT: parking_lot::ReentrantMutex<RefCell<Option<Context>>> =
    parking_lot::const_reentrant_mutex(RefCell::new(None));
```

However,  I could not event get an example `get_context()` function to compile. Instead, every call to `get_context()` would need to be replaced with "unwrapping" three layers of types before getting to the inner `Context`.
In addition, `RefCell` would add extra run-time checks to get interior mutability in addition to the `Mutex` locking and `Option` unwrapping overhead.

# Solution: Pass `Context` as parameter

Instead of using a global variable, I think the way forward is by constructing the `Context` in the true `main()` function and passing it to other functions as a parameter.

[Passing `Context` as a parameter playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4fcb98008599b334f547d33d16457b7d)

PR #368, opened 2022-01-12
[WIP] Use Context as param instead of global

tmfink:
Initial work to access `Context` struct by passing it as a parameter
instead of a global. This is an effort to improve the soundness of
macroquad (#333) and reduce maintenance burden.

This PR is a work-in-progress.

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -1,6 +1,6 @@
 //! Loading and playing sounds.
 
-use crate::{file::load_file, get_context};
+use crate::file::load_file;
 use std::collections::HashMap;
 
 #[cfg(all(feature = "audio"))]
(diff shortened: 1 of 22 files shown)


not-fl3:
Impressive work indeed! 

I am not exactly sold on the explicit context in all the macroquad API tho - I tried it with current macroquad examples and it doesn't look good.

tmfink:
> Impressive work indeed!

Thanks!

> I am not exactly sold on the explicit context in all the macroquad API tho - I tried it with current macroquad examples and it doesn't look good.

Completely implementing this change would affect basically affect all internal and user-facing APIs (not ideal given the human time it takes to implement).

However, it seems that [`ggez`](https://github.com/ggez/ggez) takes a [similar approach to what I proposed](https://github.com/ggez/ggez/blob/50918f133785e1ee585c022d94a0176ea0e73887/examples/01_super_simple.rs):

- Create context in `main()` function:
https://github.com/ggez/ggez/blob/50918f133785e1ee585c022d94a0176ea0e73887/examples/01_super_simple.rs#L45-L48
- User code gets `&mut Context` (via a trait impl):
https://github.com/ggez/ggez/blob/50918f133785e1ee585c022d94a0176ea0e73887/examples/01_super_simple.rs#L20-L42

Using this approach, `ggez` uses very little `unsafe`:
```
$ rg unsafe
src/lib.rs
181:// This is not as strong a constraint as `#![forbid(unsafe_code)]` but is good enough.
182:// It means the only place we use unsafe is then in the modules noted as allowing it.
183:#![deny(unsafe_code)]

src/graphics/mod.rs
18:#![allow(unsafe_code)]

src/graphics/shader.rs
6:#![allow(unsafe_code)]
40:unsafe impl Pod for EmptyConst {}
```

The only other approach I can think of that is sound **and** (mostly) preserves macroquad's current API is by introducing a new kind of concurrency primitive that may have a mix of:
- minor inefficiencies
- creating the possibility for deadlocks or panics

Issue #384, opened 2022-02-12
Make Circle's field public


tynberry:
Make Circle's field public to make it behave the same as Rect.

an-dr:
Have to be closed since #387 did the thing

PR #387, opened 2022-02-17
Updated Circle struct to allow for clone/copy/debug and make fields public

brendanu:
* Updated Circle struct to make x, y and r public which addresses #384
* Added in Clone, Copy, Debug, Default, PartialEq based on Rect as needed to be able to copy/clone the struct and went for matching Rect.


not-fl3:
Thanks!

PR #392, opened 2022-03-05
Add `selected` to `Button`

jackfranklin:
This PR adds a `selected` method to the UI Button, such that the
selected styling in the skin can be applied when the button is active.

Please shout if this isn't done as you'd like - I'm very new to Rust!


not-fl3:
Looks good to me! I am not really sure that having control over `selected` field will do much in terms of skinning etc, but if it works for you - then it works! :)

jackfranklin:
Thanks! I was able to style this correctly using the color_selected custom
skin builder method 👍

PR #404, opened 2022-04-11
prevent getting stuck on right edge of moving solids

adambiltcliffe:
At the moment behaviour is different when a moving solid body pushes an actor left compared to when pushing right. I don't think the -1 should be there and removing it produces consistent behaviour between the two directions.


not-fl3:
Thanks, glad it worked!

Issue #397, opened 2022-03-15
Regression in macroquad 0.3.15


sonicrules1234:
Causes android apps to crash on start.

FATAL EXCEPTION: main
Process: rust.numrows, PID: 24926
java.lang.UnsatisfiedLinkError: Unable to load native library "/data/app/~~9lW8IaHe53jHg0clZ-AJ3w==/rust.numrows-mcx-DvtS7b5xhJRH5xtwQA==/lib/arm64/libnumrows.so": dlopen failed: cannot locate symbol "glAttachShader" referenced by "/data/app/~~9lW8IaHe53jHg0clZ-AJ3w==/rust.numrows-mcx-DvtS7b5xhJRH5xtwQA==/lib/arm64/libnumrows.so"...
	at android.app.NativeActivity.onCreate(NativeActivity.java:178)
	at android.app.Activity.performCreate(Activity.java:7994)
	at android.app.Activity.performCreate(Activity.java:7978)
	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1309)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3404)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3595)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:223)
	at android.app.ActivityThread.main(ActivityThread.java:7664)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)

PR #409, opened 2022-05-07
Fix android builds by specifying miniquad's commit

Yeicor:
This is a workaround for https://github.com/not-fl3/macroquad/issues/397.

If instead of specifying miniquad's version, I reference **the same commit** of that version, the build for Android works (using the docker image). I don't know why, maybe It is related to the cargo version of the container (1.57.0).

The build command I use is `docker run -it --rm -v $(pwd)":/root/src" -w /root/src notfl3/cargo-apk bash -c "rm -rf /usr/local/cargo/registry && cargo quad-apk build"`

Is this an acceptable workaround to be able to use the latest macroquad on Android?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ opt-level = 3
 all-features = true
 
 [dependencies]
-miniquad = { version = "=0.3.0-alpha.46", features = ["log-impl"] }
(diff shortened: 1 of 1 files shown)


Yeicor:
I realized that this can also be applied without accepting the pull request by using the following in user projects (`Cargo.toml`):

```toml
[patch.crates-io]
miniquad = { git = "https://github.com/not-fl3/miniquad", rev = "8ad337af4b53f9e65deccf2fd9681936a344b30a" }
```

However, It still does not work with egui-macroquad, so a better fix is needed :pensive:.

PR #413, opened 2022-05-26
Allows coroutines to return values

nobbele:
with this you can return values from coroutines like this
```rust
#[macroquad::test]
async fn coroutine_value() {
    let mut coroutine = start_coroutine(async move {
        next_frame().await;
        1
    });

    coroutine.set_manual_poll();

    assert_eq!(coroutine.retreive(), None);

    coroutine.poll(0.0);
    coroutine.poll(0.0);

    assert_eq!(coroutine.retreive(), Some(1));
}
```

### DIFF
--- a/src/experimental/coroutines.rs
+++ b/src/experimental/coroutines.rs
@@ -3,7 +3,9 @@
 //! some evaluation over time.
 //!
 
+use std::any::Any;
 use std::future::Future;
+use std::marker::PhantomData;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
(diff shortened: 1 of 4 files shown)


src/experimental/coroutines.rs
```diff
@@ -33,24 +62,38 @@ impl CoroutinesContext {
...
+impl<T: 'static> Coroutine<T> {
+    /// Returns true if the coroutine finished or was stopped.
     pub fn is_done(&self) -> bool {
         let context = &get_context().coroutines_context;
 
-        context.coroutines[self.id].is_none()
+        context.coroutines[self.id].is_value() || context.coroutines[self.id].is_nothing()
+    }
+
+    pub fn retreive(self) -> Option<T> {
```
not-fl3, line 89: isnt it "retrieve"?

src/experimental/coroutines.rs
```diff
@@ -93,32 +136,35 @@ impl Coroutine {
...
         let coroutine = &mut context.coroutines[self.id];
-        if let Some(f) = coroutine {
+        if let CoroutineState::Running(f) = coroutine {
             context.active_coroutine_now = f.manual_time;
             context.active_coroutine_delta = Some(delta_time);
             *f.manual_time.as_mut().unwrap() += delta_time;
-            if resume(&mut f.future) {
-                *coroutine = None;
+            if let Some(v) = resume(&mut f.future) {
+                *coroutine = CoroutineState::Value(v);
```
not-fl3, line 144: With current implementation it is not really important if stopped/finished coroutine have `CoroutineState::Value(())` or `CoroutineState::Nothing` value, so this should not be addressed right now

but in the future, when coruitne's id's will be reused to stop leaking memory - it would be nice to somehow automatically auto-retrieve values of finished `()` coroutines and therefore change the state from `::Value(())` to `Nothing`

not-fl3:
Great work, thanks for PR!

And you're right, probably it's time to fix the memory leak and start reusing memory from finished coroutines! Sure out of the scope of this PR :)

src/experimental/coroutines.rs
nobbele: I thought about adding a bool in `CoroutineInternal` that gets set when all `Coroutine` copies get dropped (using arc), it would require removing `Copy` from `Coroutine`.

PR #414, opened 2022-05-27
coroutines: Fix memory leak

not-fl3:
Now coroutines reuse memory from older, finished coroutines and therefore starting new coroutines doesn't leek the memory.

There is still a way to leak the memory, though - if coroutine return the value and that value is never retrieved - macroquad will keep this value forever. The only fix available - make coroutines non-copy and mark them the value "unwanted"  in the Drop.

Not too much of an issue - the intention of returning a value from a coroutine - is always to retrieve it (hopefully).

cc @nobbele

### DIFF
--- a/src/experimental/coroutines.rs
+++ b/src/experimental/coroutines.rs
@@ -3,7 +3,7 @@
 //! some evaluation over time.
 //!
 
-use std::any::Any;
+use std::any::{Any, TypeId};
 use std::future::Future;
 use std::marker::PhantomData;
 use std::pin::Pin;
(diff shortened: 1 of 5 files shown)


src/experimental/coroutines.rs
```diff
@@ -46,51 +54,83 @@ impl CoroutineState {
...
+        return true;
     }
 
-    pub fn retrieve(self) -> Option<T> {
+    pub fn retrieve(&self) -> Option<T> {
         let context = &mut get_context().coroutines_context;
 
-        if let Some(v) = context.coroutines[self.id].take_value() {
-            return Some(*v.downcast().unwrap());
+        // () is a special case. Futures with () as a return type do not keep
```
nobbele, line 121: This special casing shouldn't be necessary? a `Coroutine<()>` will still produce a `Box<()> as Box<dyn Any>` so the general case should work fine. Except that this special case will always return Some(()), even multiple times, is that intentional?

nobbele: nevermind I see the reason now

Issue #393, opened 2022-03-06
Feature Request: Rotated text


e00E:
draw_texture allows rotation but draw_text does not.

an-dr:
Also can be closed

PR #415, opened 2022-05-27
Add rotation to draw_text_ex method

uggla:
This PR fix issue  #393.

* Add rotation parameter to draw_text_ex.
* Add a rotating text into text example.

Note: I think adding a new method to retrieve the text center could be convenient. I will try to add one in a next PR unless you want it in this one.

### DIFF
--- a/examples/text.rs
+++ b/examples/text.rs
@@ -6,6 +6,8 @@ async fn main() {
         .await
         .unwrap();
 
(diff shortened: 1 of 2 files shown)


not-fl3:
Thansk for PR, really sorry for the late merge!

uggla:
@not-fl3, no worries. Thanks merging it.

PR #417, opened 2022-06-08
Remove Collisions With Out-of-Bounds Tiles

zicklag:
A fix for https://github.com/fishfight/FishFight/issues/135.

I've only tested it with FishFight, but it appeared to fix the problem.

### DIFF
--- a/physics-platformer/src/lib.rs
+++ b/physics-platformer/src/lib.rs
@@ -368,11 +368,16 @@ impl World {
             tag: layer_tag,
         } in &self.static_tiled_layers
         {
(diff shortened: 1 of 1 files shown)


not-fl3:
Goot catch, thanks!

zicklag:
@not-fl3 would it be possible to get a new Crates.io release for `macroquad-platformer` with this change?

not-fl3:
> @not-fl3 would it be possible to get a new Crates.io release for `macroquad-platformer` with this change?

Sure, `0.1.3` published

zicklag:
Awesome, thanks!

PR #433, opened 2022-07-11
math: Rect::center

TanTanDev:
I think a center() function for the Rect, could be useful in many cases.
then we don't have to type this every time:
`let center = rect.point() + rect.size() * 0.5f32;`


not-fl3:
Nice, thanks!

PR #435, opened 2022-07-14
time: `Timer` struct

ryangjchandler:
I've found myself hand-rolling timers a couple of times. This pull request implements a basic implementation that requires "ticking" the timer each frame.


not-fl3:
Hi! The `Timer` looks pretty useful! 
However, I am not really sure if it should be in macroquad itself or in a crate, maybe with some other goodies to simplify life with macroquad?

PR #437, opened 2022-07-16
Fix AtlasConfig wrong matrix index calculation

vdrn:
When calculating matrix index from Particle AtlasConfig, to get row index, array index needs to be divided by width, not height


not-fl3:
wow, good catch, thanks for PR!

PR #443, opened 2022-07-18
Add several functions to interact with animations using strings

DaQueenJodi:
I noticed there was a `name`  field on the `Animation` struct, so I figured it would be useful to use it to set the current animation, get the name of the current animation, and be able to get the index from the name
i wasn't sure what to call the functions, but hopefully these are fine.

### DIFF
--- a/src/experimental/animation.rs
+++ b/src/experimental/animation.rs
@@ -1,6 +1,6 @@
 use crate::{
-    math::{vec2, Rect, Vec2},
-    time::get_frame_time,
+    math::{vec2, Rect, Vec2}, prelude::get_frame_time,
+    
 };
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


DaQueenJodi:
it may be better to take a `&str` for each function, which should be pretty easy to change
I decided to use a `String` for all of them since `get_current_str()` would need to return a `String` and not a `&str`, but that may not be that important

src/experimental/animation.rs
```diff
@@ -46,6 +47,22 @@ impl AnimatedSprite {
...
     }
 
+    pub fn get_animation_index(&self, name: String) -> Option<usize> {
+        self.animations.iter().position(|animation| animation.name == name)
+    }
+
+    pub fn set_animation_str(&mut self, name: String) {
+        self.current_animation = self.get_animation_index(name.clone()).expect(&format!("{} is not a valid animation for this sprite!", name));
+    }
+    pub fn current_animation_str(&self) -> String {
```
not-fl3, line 57: I believe returning String is totally fine

src/experimental/animation.rs
```diff
@@ -46,6 +47,22 @@ impl AnimatedSprite {
         }
     }
 
+    pub fn get_animation_index(&self, name: String) -> Option<usize> {
```
not-fl3, line 50: but as an argument `&str` is a better choice

DaQueenJodi:
I also added a way to get the number of frames in the current animation, which could be useful, would you want me to commit that to the PR as well?

PR #444, opened 2022-07-21
Particles: fix ParticleShape::Rectangle and add rotation params to EmmiterConfig

vdrn:
These 2 commits can be applied independently (sorry for batching them into one pr)

#### e8d9714 **Add aspect_ratio for ParticleShape::Rectangle**
Currently, EmmiterConfig's `shape` field can be set to `EmmisionShape::Rectangle`, but what you get is a square.
Adding `aspect_ratio` seems like the simplest solution to have rectangles.

Since this commit modifies rectangle shape to `ParticleShape::Rectangle{aspect_ratio: f32}` it is a **breaking change** for those who explicitly set `shape: ParticleShape::Rectangle`. Hopefully, this is a rare case since it is the default setting.

#### e5bc288 **Add particle rotation parameters**
This adds the following params to EmitterConfig:
```
initial_rotation: f32,
initial_rotation_randomness: f32,
initial_angular_velocity: f32,
initial_angular_velocity_randomness: f32,
angular_accel: f32,
angular_damping: f32,
```
To quickly hack this, I've used the z coordinate of `GpuParticle.pos` to pass rotation to the vertex shader.
This coordinate is currently unused, but might not be the best solution.

Edit: added a third commit that fixes wrong initial rotation

### DIFF
--- a/particles/src/lib.rs
+++ b/particles/src/lib.rs
@@ -193,6 +193,21 @@ pub struct EmitterConfig {
     /// Velocity acceleration applied to each particle in the direction of motion.
     pub linear_accel: f32,
 
+    // Initial rotation for each emitted particle.
+    pub initial_rotation: f32,
+    /// Initial rotation randomness.
+    /// Each particle will spawned with "initial_rotation = initial_rotation - initial_rotation * rand::gen_range(0.0, initial_rotation_randomness)".
+    pub initial_rotation_randomness: f32,
+    // Initial rotational speed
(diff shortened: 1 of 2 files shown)


not-fl3:
Wow! 

Thanks for PR! 

just FYI, this breaks the [editor](https://github.com/not-fl3/particles-editor) a bit, but no worries about that, new params are always worth it.

vdrn:
You're welcome, I'm glad to be of help!

PR #446, opened 2022-07-23
Update to 2021 edition

e00E:
I didn't see a reason to stay on the old edition so we should give people working on this project the advantages of the new edition.


svenstaro:
@not-fl3 How about just merging this? Seems low risk.

not-fl3:
oh, sure, done!

PR #447, opened 2022-07-23
Tiled: optimize draw calls batching

vdrn:
#### Problem:
When drawing a tile layer, macroquad tries to batch draw calls, but batching is reset every time the tile tileset changes.
This means that for drawing a single layer, we need `number_of_tilesets ..= number_of_tiles` draw calls (eg. if all odd tiles are from one tileset, and even from a different one, we will have a draw call for each tile).

#### Solution:
Before drawing tiles, we separate them by tileset. This enables macroquad to batch all tiles from one tileset together, so we will always have `number_of_tilesets` draw calls.

### DIFF
--- a/tiled/src/lib.rs
+++ b/tiled/src/lib.rs
@@ -144,20 +144,30 @@ impl Map {
         let spr_width = dest.w / source.w;
         let spr_height = dest.h / source.h;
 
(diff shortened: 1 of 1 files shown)


not-fl3:
Nice, thanks for PR!

PR #448, opened 2022-07-25
Add draw_rectangle_ex function and example for using it

vdrn:
This function merges functionalities of `draw_rectangle` and `draw_rectangle_lines`, and adds:
- rotation with pivot
- rounded edges
- horizontal and vertical skewing
- 4 color gradient fill

PS. also fixed error in docs for `draw_texture_ex`. Not sure what's the best wording, but pivot is in same coordinate system as `x` and `y` args.


musjj:
@not-fl3 
Hey can you take a look at this PR? Round-cornered rectangles would be nice to have.

not-fl3:
Oh,  sorry for delay! On it!

PR #458, opened 2022-08-06
Limit particle lived timer

vdrn:
If `lived` is not limed to `lifetime`, condition on line [835](https://github.com/not-fl3/macroquad/blob/b37e40305e6eda6848d57077e769c9f558dd0af5/particles/src/lib.rs#L835) can be true, which in turn causes emitting of new, unwanted, particles.

This happens for example if `explosiveness: 1.0, lifetime_randomness: 0.5`


not-fl3:
Thanks! :+1:

PR #460, opened 2022-08-06
Fix panic when drawing out of bound tiles

vdrn:
This line (which I wrote :) ) panics when user passes source rectangle that is out of bounds


not-fl3:
Thanks!

PR #461, opened 2022-08-06
Fix particle's lifetime

vdrn:
Limiting `lived` introduced a new problem: particles were not removed until the end of `config.lifetime` (instead of their individual lifetimes)


not-fl3:
:heart: 

thanks for all your PRs on particles!

PR #473, opened 2022-08-19
draw_texture_ex: Fix source rectangle when using batched texture

vdrn:
This fixes source rect calculation when using batched texture with custom source rect.

That in turn enables texture batching to work with macroquad featured like AnimatedSprite od Tiled drawing


not-fl3:
Wow, thanks for PR!

PR #484, opened 2022-08-29
Calculate Circle.overlaps_rect from shape center

### LINKED ISSUE #483: `Circle.overlaps_rect` is not calculated correctly
The current implementation assumes that `x` and `y` of the comparing `Rect` are centered. They are in fact the leftmost and topmost points of the rect, causing the overlap to be off by half the rect's `width` and `height`.

roqvist:
fixes #483


not-fl3:
Thanks!

PR #485, opened 2022-09-01
Fix particle blending

### LINKED ISSUE #472: Wrong particle blending when using RenderTarget
Rendering to `RenderTarget` and drawing particles using default `BlendMode::Alpha` produces unexpected results.
Particles seem to be blended first with black color, and then with screen texture.
When not using `RenderTarget`, particles are correctly blended just with screen texture.

Here's a gist that reproduces both situations:
https://gist.github.com/vdrn/933d52a32191ca1b5c0b9fb15b1e02ca

vdrn:
Fixes #472

### DIFF
--- a/particles/src/lib.rs
+++ b/particles/src/lib.rs
@@ -563,6 +563,11 @@ impl Emitter {
             shader,
             PipelineParams {
                 color_blend: Some(blend_mode),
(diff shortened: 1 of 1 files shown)


not-fl3:
Thank you! :+1:

PR #488, opened 2022-09-08
Fix font size scaling in `draw_text_ex`

### LINKED ISSUE #487: Font sizes different between WASM/Native
I'm not sure if I'm just not read up enough on how font sizing should work, but the following code:

```rust
use macroquad::prelude::*;

#[macroquad::main("Foo")]
async fn main() {
    request_new_screen_size(400.0, 400.0);

    loop {
        draw_rectangle(0.0, 0.0, 200.0, 200.0, WHITE);
        draw_text("Testing 123", 10.0, 100.0, 32.0, BLACK);

        next_frame().await;
    }
}
```

produces different font sizes on WASM vs Windows.

Windows:
font_native

Firefox:
font_wasm

Reproduction repo:
https://github.com/rileylyman/font_size_repro

rileylyman:
Fixes #487, not sure if this is the right change but it seems like this is more correct.

After this change both Firefox and Windows display the same font size when the reproduction repo is run.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -287,7 +287,7 @@ pub fn draw_text_ex(text: &str, x: f32, y: f32, params: TextParams) {
     let font_scale_y = params.font_scale;
     let dpi_scaling = get_quad_context().dpi_scale();
 
-    let font_size = params.font_size * dpi_scaling.ceil() as u16;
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks!

Issue #459, opened 2022-08-06
Platformer example falling through bottom tiles


vdrn:
Bottom tiles (the ones under the moving collider) do not have collision.
I see that they do in the web example, but when building native (on Linux) character falls through.

Reverting this [commit](https://github.com/not-fl3/macroquad/commit/a76a2dc001f1d752d5b0689c8382bfd54fcf69ea) fixes the issue

nezvers:
I did a quick test and drew red squares on solid tiles, those floor tiles are solid. My guess is that there's a miscalculation and doesn't take into account the last row.

13-05:
hey - i think this issue is still present in macroquad? if i clone this repo and cd into it, then do `cargo run --example platformer`, the lowest level of the floor is solid. but if i copy all the code+assets to a new Rust project with the assets stored in an `assets` subdirectory of the project and the `Cargo.toml`'s dependencies looking like this:
```toml
[dependencies]
macroquad = "0.3.25"
macroquad-tiled = "0.1.1"
macroquad-platformer = "0.1.3"
```

it does this:
![Peek 2022-12-29 18-27](https://user-images.githubusercontent.com/64506392/210023478-a85eec5d-391f-4dc3-adb0-0a3be11eb114.gif)

that is, the character still falls through the lowest floor level.

PR #498, opened 2022-09-23
FIX: falling trough lowest row of solids

nezvers:
After trying out examples I noticed a bug (#459 ) in the platformer example.
In mentioned issue was mentioned latest code change when the bug appeared. My guess was that the last row of tiles is not taken into an account and increasing layer_height by one fixed it.


not-fl3:
Good catch, thanks!

PR #500, opened 2022-09-27
particles: fix particles lifetime

not-fl3:
This change https://github.com/not-fl3/macroquad/pull/458 introduced a regression: `examples/particles_example` got broken.

I am not quite sure how it should be properly fix, so for now I'll just rollback the change to get the example alive again. 

I guess something need to be done with https://github.com/not-fl3/macroquad/pull/461 as well? 

cc @vdrn

### DIFF
--- a/particles/src/lib.rs
+++ b/particles/src/lib.rs
@@ -808,8 +808,8 @@ impl Emitter {
                 gpu.data.y = cpu.lived / cpu.lifetime;
             }
 
-            cpu.lived = f32::min(cpu.lived + dt, cpu.lifetime);
-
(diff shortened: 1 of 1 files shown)


vdrn:
Hi @not-fl3, sorry for the trouble.

I'm not sure either what would be the best way forward.

After having a quick look, this also seems to fix the issue: https://github.com/vdrn/macroquad/commit/10f1dcdce29add4944f255fc710bda9e9fbfd62c

I'm just leaving it here as it might be helpful for future fixes. Unfortunately, I'm not using this code directly in my project anymore (I've copied it and used as a base for some custom stuff I'm doing).

PR #501, opened 2022-10-01
Propagate error from when loading textures or images instead of panicking

actuday6418:
At line 209 and 216, it's still using `expect` instead of propagating the error further. Just not sure what would be the best way to do that. I can think of using `map_err` to convert the error type with relevant error message, Or creating a new `Err` type `enum` with `ImageError` and `FileError` variants.

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -9,6 +9,7 @@ use crate::{
 
 use crate::quad_gl::{DrawMode, Vertex};
 use glam::{vec2, Vec2};
+use image::ImageResult;
 
 pub use crate::quad_gl::FilterMode;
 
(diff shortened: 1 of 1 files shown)


src/texture.rs
```diff
@@ -206,14 +206,14 @@ impl Image {
 pub async fn load_image(path: &str) -> Result<Image, FileError> {
     let bytes = load_file(path).await?;
 
-    Ok(Image::from_file_with_format(&bytes, None))
+    Ok(Image::from_file_with_format(&bytes, None).expect("Valid image required"))
```
not-fl3, line 209: shouldn't it propogate the error as well?

actuday6418: Yes it should, but `from_file_with_format` returns an `ImageError` in it's `Result`, while the current function takes `FileError`. I'm not sure what's the way to handle this (what I usually do is create an new `enum` type with both errors, or just use `anyhow`)

not-fl3: I believe to justify the breaking change in the API the errors thing should be solved, I do not want to break the API twice - with this PR and with the future PR fixing this panic

actuday6418: Fixed in latest commit. Could you review?

not-fl3: Looks good to me! Could you remove irrelevant commits from this PR? And I am ready to merge!

actuday6418: Removed unrelated change in Cargo.toml

not-fl3: now just ui_skins left :)

actuday6418: The changes in ui_skins is necessary to pass tests right? Or should that be in another PR?

PR #504, opened 2022-10-12
derive(Debug, Clone, Copy) for TextDimensions

rope-hmg:
Ran the tests. All of the coroutine and back to the future tests segfaulted, but they were doing that after a fresh clone so I'm assuming they're supposed to.


not-fl3:
thanks for PR!

> All of the coroutine and back to the future tests segfaulted

Might be worth opening an issue with some details!

PR #506, opened 2022-10-22
Implemented Hexadecimal to RGBA color conversion

eboatwright:
- Added hexadecimal.rs file with: hex_to_u8 and hex_pair_to_u8 functions
 - added Color::from_hex that converts a hexadecimal str in the format of "#RRGGBB" or "#RRGGBBAA"

 - I know the code isn't the best, but I've wanted this function for a while


Aplosion:
Wrote a similar function in a program I was working on:

```//uses a u32 to store the hex value, so that hex codes can be copied directly into my code.
//wastes the upper four bytes, but I think this is fine, any other way I could think of to do this
//didn't have the convenience and features I wanted. And it's much faster than messing with strings.
//can be expanded to support RGBA with the wasted bytes, I guess.
fn hex_to_float(a: u32) -> Color {
    let u32_as_bytes: [u8; 4] = a.to_be_bytes();
    Color::new(
        u32_as_bytes[1] as f32 / 255.0,
        u32_as_bytes[2] as f32 / 255.0,
        u32_as_bytes[3] as f32 / 255.0,
        1.0,
    )
}
```

Aplosion:
Go for it! Glad you like my code :)

⁣Get BlueMail for Android ​

eboatwright:
Yeah, thanks so much! That string manipulation code took me a whole afternoon to put together, and this system is just so much cleaner

not-fl3:
Thanks for PR, sorry for the late merge!

eboatwright:
:D

PR #513, opened 2022-11-07
Fix macroquad not building under wasm when default features are disabled

FaunKr:
This fixes https://github.com/not-fl3/miniquad/issues/317

The build failure was caused by the missing `is_loaded` function in the dummy audio implementation. 
I'm not sure if returning `true` would be a better choice in case a project decides to wait for it to return true. 

I also added `delete` to prevent similiar problems in the future.


not-fl3:
Thanks!

PR #514, opened 2022-11-08
Update Cargo.toml

dimiour:
image crate requires "extern_crate_alloc" feature


dimiour:
My bad, I think this is an issue with the image crate. Wondering if you are experiencing the same issue?

image

I am compiling a blank macroquad project. I tried compiling the image crate as well and it returns the same error.

not-fl3:
It looks like an outdated rustc version? Feel free to reopen the issue if the problem persists

PR #516, opened 2022-11-12
Make load_ttf_font failures recoverable

FaunKr:
So far `load_ttf_font` would panic when called on a non-existing file. This seemed counter intuitive to me as it returns a Result. 

So i changed it to return an error if it fails to load the file. 

Right now it still can crash when being called under wasm, but this should be resolved with this https://github.com/not-fl3/miniquad/pull/323 merge request, and then the behavior should be consistent across all plattforms.


not-fl3:
Nice, thanks for PR!

PR #517, opened 2022-11-15
Added `draw_rectangle_pro`

Tekkunsan:
## What?
I added a function that renders a rectangle with a position (centered), a rotation (degrees), and a size

## Why?
Right now, the only way to rotate a quad is using `draw_poly` so I created this function

## How?
I created a transformation matrix and multiply it to the vertices

## Example
rotating

### DIFF
--- a/src/shapes.rs
+++ b/src/shapes.rs
@@ -74,8 +74,8 @@ pub fn draw_rectangle_lines(x: f32, y: f32, w: f32, h: f32, thickness: f32, colo
     context.gl.geometry(&vertices, &indices);
 }
 
-/// Draws a solid rectangle with its center at `position` with `size` (width going to
-/// the right, height going down), with a given `color`.
+/// Draws a solid rectangle with its center at `position` with `size`,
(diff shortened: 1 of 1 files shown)


src/shapes.rs
```diff
@@ -74,6 +74,36 @@ pub fn draw_rectangle_lines(x: f32, y: f32, w: f32, h: f32, thickness: f32, colo
     context.gl.geometry(&vertices, &indices);
 }
 
+/// Draws a solid rectangle with its center at `position` with `size`,
+/// with rotation (in degrees), with a given `color`.
+pub fn draw_rectangle_pro(position: Vec2, rotation: f32, size: Vec2, color: Color) {
```
not-fl3, line 79: for "optional arguments" macroquad use `do_whatever_ex` + a struct with arguments notation

you can check `draw_texture_ex` as an example

Tekkunsan:
I also made it so that you use the `offset` param to offset the position instead of just it being in the center

not-fl3:
Nice, thanks for PR!

PR #521, opened 2022-11-22
Add draw_pretty_circle that draws a prettier circle than draw_circle

Jakkestt:
I have no idea what im doing. Do not judge.


not-fl3:
Hi! My first response - maybe this deserves a separate crate?

Jakkestt:
In what way?

Issue #349, opened 2021-12-25
input_text doesn't show text correctly when display scaling is enabled


vpzomtrrfrt:
Similar conditions to #331

When display scaling is enabled, input fields (and maybe other UI elements too?) appear to cut off the contained text

```rust
use macroquad::prelude::*;
use macroquad::ui::{hash, root_ui};

#[macroquad::main("inputtest")]
async fn main() {
    let mut value = String::new();

    loop {
        clear_background(GREEN);

        root_ui().input_text(hash!(), "Input", &mut value);

        next_frame().await;
    }
}
```

Without display scaling:
![bildo](https://user-images.githubusercontent.com/3528358/147378218-c74c8d44-0ac6-408d-a475-e3bc367d64d8.png)

With display scaling:
![bildo](https://user-images.githubusercontent.com/3528358/147378227-40e3fc9c-ec33-43e1-a791-1cb397d4820a.png)

vpzomtrrfrt:
Also reproducible with Editbox directly

vpzomtrrfrt:
This patch appears to fix it, though I'm not sure if it's the right place:
```patch
diff --git a/src/ui/render/painter.rs b/src/ui/render/painter.rs
index 2167bd3..886b3bf 100644
--- a/src/ui/render/painter.rs
+++ b/src/ui/render/painter.rs
@@ -497,6 +497,11 @@ impl Painter {
         let rect = rect.into();
 
         self.clipping_zone = if let Some(rect) = rect {
+            let dpi = crate::get_context().quad_context.dpi_scale();
+
+            let mut rect = rect;
+            rect.scale(dpi, dpi);
+
             Some(self.clipping_zone.and_then(|old_rect| old_rect.intersect(rect)).unwrap_or(rect))
         } else {
             None
```

vpzomtrrfrt:
hmm, perhaps I spoke too soon, doesn't seem like that changed the behaviour in my actual code

vpzomtrrfrt:
The amount that is obscured changes when the window is resized

vpzomtrrfrt:
Updated patch to only change the clipping in the draw command itself:
```patch
diff --git a/src/ui/render/painter.rs b/src/ui/render/painter.rs
index 2167bd3..5f10641 100644
--- a/src/ui/render/painter.rs
+++ b/src/ui/render/painter.rs
@@ -502,8 +502,13 @@ impl Painter {
             None
         };
 
[...]

PR #522, opened 2022-11-26
Fixing clipping and offset issues on high DPI displays

### LINKED ISSUE #464: Incorrect UI rendering on MacOS built-in Retina display
I just copied the [UI example code](https://github.com/not-fl3/macroquad/blob/master/examples/ui.rs) and executed it.

The UI example looks fine both on my Linux machine, and on my MacBook Pro (2019, the ones with the touch bar) as long as I show it on my lower-resolution external monitor (2560x1440):
macroquad-ui-ok

However, when the UI example is on the MacBooks own display (the Retina display), it looks like this:
macroquad-ui-broken

Note that the contents of all three windows is gone (and the window titles' positions are slightly off).

canselcik:
Seems to fix #464 and #349

### DIFF
--- a/src/ui/render/painter.rs
+++ b/src/ui/render/painter.rs
@@ -272,7 +272,7 @@ impl Painter {
 
                 let left_coord = (element_size.x - text_measures.width) / 2.;
                 let top_coord =
-                    element_size.y / 2. - text_measures.height / 2. + text_measures.offset_y;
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks!

afischl:
Thank you very much, @canselcik ! This has been a significant improvement but the offsets are still slightly wrong as I've described in this new ticket: #530

PR #526, opened 2022-12-07
show/hide android virtual keyboard on input focus/blur

macnelly:
This shows android virtual keyboard when an input has focus.
If the input blurs the keyboard will get hidden again.

Or is there any reason this is not already implemented? There is no input focus event (I am aware of) so this cannot be done from outside.


not-fl3:
Thanks!

PR #527, opened 2022-12-07
conditional import of get_quad_context

macnelly:
sorry:
made the android only used import conditional to overcome compiler warning


not-fl3:
`cargo fmt` doesn't look very happy :)

macnelly:
I was not happy for audio.rs, text.rs, painter.rs and my touched file - fixed :)

PR #532, opened 2022-12-10
Fix two bugs in particles, one of which causes performance issue

Mivik:
The first commit fixes a performance bug in the place where particles out of their lifetimes are removed. The original version uses reversed iterators and `remove` function. When there are `m` out of `n` particles should be removed, the expected time complexity would be `O(nm)` which is unaffordable. Simply replacing `remove` with `swap_remove` solves the problem since we iterate backwards.

The second commit fixes the problem that the viewport is not taken into consideration when setting up render pass. I'm not too familiar with these graphics stuff but I think it's a mistake.


Mivik:
For me the fix improves the fps (1s / draw_time) from 60 to 400 under extreme situation.

PR #535, opened 2022-12-14
texture: update_part

troyc:


not-fl3:
LGTM, but could you check why `cargo fmt` is unhappy?

not-fl3:
> > LGTM, but could you check why `cargo fmt` is unhappy?
> 
> I ran `cargo fmt` and amended the commit.

perfect, thanks for PR! :+1:

PR #538, opened 2022-12-15
Add animation documentation

Mifom:
The animation API looks pretty good but lacking documentation.

This PR trying to fix this.


not-fl3:
thanks!

PR #540, opened 2022-12-19
Further fix label alignment on high-dpi displays

### LINKED ISSUE #530: Still incorrect UI rendering on high-dpi display
This is a follow-up to #464 and #522

The code change massively improved the situation (as described in #464), but unfortunately we're not quite there yet.

I'm using the [UI example code](https://github.com/not-fl3/macroquad/blob/master/examples/ui.rs) again to illustrate the situation. Please look at the screenshot below and note these two problems in the window in the foreground (top right corner):

- When expanding the editbox, the contained text _This is editbox!_ overlaps with the editbox' label above
- The plus/minus icons are slightly off to the top and this continuously adds up (it's easier to spot further down). This doesn't look much but I clicked the wrong plus button several times, more often than the difference on the screenshot might suggest 😄 

Screenshot_MacOS

There's no new release on crates.io, so I've built [commit 5c5e8f4](https://github.com/not-fl3/macroquad/commit/5c5e8f41caf05f3594d7c1e151bea7a85114c185) locally.

canselcik:
This takes care of the issue on TreeNode and probably in other places as well. It will fix #530.

I think it is a good idea to merge this like the last occurrence but also looks like it is worth exploring if `offset_y` in `TextMeasures` should be DPI aware instead.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -166,7 +166,7 @@ impl FontInternal {
         TextDimensions {
             width: width / dpi_scaling,
             height: height / dpi_scaling,
-            offset_y: max_y,
+            offset_y: max_y / dpi_scaling,
         }
     }
 }
--- a/src/ui/render/painter.rs
+++ b/src/ui/render/painter.rs
@@ -239,6 +239,7 @@ impl Painter {
         let background_margin = style.background_margin.unwrap_or_default();
         let margin = style.margin.unwrap_or_default();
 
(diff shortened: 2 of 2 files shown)


canselcik:
I have explored the proposal I had earlier. This should address the issue closer to its source.

not-fl3:
Nice! 
LGTM apart of `cargo fmt` being a bit unhappy

PR #556, opened 2023-01-31
Fix DPI on camera viewport

piemanau:
## Whats the issue?
The issue is that the DPI scaling is different when rendering to the normal camera compared to when rendering to a custom made camera. I just tweaked the camera viewport to fix this issue. I tweaked x, y, width and height as they were all incorrect. I have included some examples to show the code I have written for the screenshots.

## Before
Screenshot 2023-01-31 at 4 19 59 pm
Screenshot 2023-01-31 at 4 20 07 pm

## After
Screenshot 2023-01-31 at 4 19 04 pm
Screenshot 2023-01-31 at 4 19 12 pm

### DIFF
--- a/src/camera.rs
+++ b/src/camera.rs
@@ -4,7 +4,7 @@ use crate::{
     get_context,
     math::Rect,
     texture::RenderTarget,
-    window::{screen_height, screen_width},
+    window::{screen_height, screen_width}, get_quad_context,
 };
 use glam::{vec2, vec3, Mat4, Vec2, Vec3};
 
@@ -106,7 +106,10 @@ impl Camera for Camera2D {
     }
 
     fn viewport(&self) -> Option<(i32, i32, i32, i32)> {
-        self.viewport
+        match self.viewport {
+            Some((x, y, w, h)) => Some((x * get_quad_context().dpi_scale() as i32, y * get_quad_context().dpi_scale() as i32, w * get_quad_context().dpi_scale() as i32, h * get_quad_context().dpi_scale() as i32)),
+            None => self.viewport,
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/camera.rs
```diff
@@ -106,7 +106,10 @@ impl Camera for Camera2D {
     }
 
     fn viewport(&self) -> Option<(i32, i32, i32, i32)> {
-        self.viewport
+        match self.viewport {
```
not-fl3, line 109:
```
let dpi_scale = ..;
self.viewport.map(|(x, y, w, h)| (x * dpi_scale, ..))
```

will look a little better :)

not-fl3: also I am not really sure if `camera.viewport` and `camera.viewport()` returning different values is the right way to fix it

piemanau: > ```
> let dpi_scale = ..;
> self.viewport.map(|(x, y, w, h)| (x * dpi_scale, ..))
> ```
> 
> will look a little better :)

I thought there would have been a better way but I wasn't quite sure what it was, I will add this now

piemanau: > also I am not really sure if `camera.viewport` and `camera.viewport()` returning different values is the right way to fix it

Possibly not, probably best to change where it is instantiated I guess

piemanau: > > ```
> > let dpi_scale = ..;
> > self.viewport.map(|(x, y, w, h)| (x * dpi_scale, ..))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will look a little better :)
> 
> I thought there would have been a better way but I wasn't quite sure what it was, I will add this now

Had to add them all individually

piemanau: > > also I am not really sure if `camera.viewport` and `camera.viewport()` returning different values is the right way to fix it
> 
> Possibly not, probably best to change where it is instantiated I guess

It is instantiated when the camera is made, so unless I'm missing something it can't be modified when made

PR #564, opened 2023-02-19
fix backspace handling in editbox widget

### LINKED ISSUE #552: Delete (Backspace) key on Mac (M1) glitches in editbox
Delete (backspace on PC) key seems glitchy on macos (M1 Air 2020) in editbox or input_text

It doesn't do anything on tap, but when held it inserts bunch of spaces and after 8 or so spaces, works normally, that means it erases those spaces and then the text as it should.

https://user-images.githubusercontent.com/43041139/211876746-d63beed6-4a7f-4d75-8083-2c0f647cc285.mov

adambiltcliffe:
On at least some platforms, pressing backspace sends a key event for the backspace character as well as a backspace event, which was causing deleting characters with the backspace key to not work properly in the editbox widget.

Fixes #552

### DIFF
--- a/src/ui/widgets/editbox.rs
+++ b/src/ui/widgets/editbox.rs
@@ -91,6 +91,7 @@ impl<'a> Editbox<'a> {
                 } => {
                     if character != 13 as char
                         && character != 10 as char
+                        && character != 8 as char
                         && character.is_ascii()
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks!!

PR #565, opened 2023-02-20
Fix backspace handling in editbox for macbooks

### LINKED ISSUE #552: Delete (Backspace) key on Mac (M1) glitches in editbox

107zxz:
Followup to #564 adding the keycode for backspacing on a Macbook, which fixes #552.

### DIFF
--- a/src/ui/widgets/editbox.rs
+++ b/src/ui/widgets/editbox.rs
@@ -92,6 +92,7 @@ impl<'a> Editbox<'a> {
                     if character != 13 as char
                         && character != 10 as char
                         && character != 8 as char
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks!

107zxz:
Sorry! Closed it by accident

PR #566, opened 2023-02-20
Replace list of keycodes to ignore in editboxes with is_ascii_control()

107zxz:
Another followup to 3b141dc. Replaces the individual keycode checks in editbox.rs with a blanket check for ascii control codes. This cleans up the previous code and also fixes the tab key behaving incorrectly in inputboxes.


not-fl3:
Thanks for PR!

PR #567, opened 2023-02-22
Add functions to draw parallelograms and parallelepipeds in 3D space

unc0mm0n:
The functions `draw_affine_parallelogram` and
`draw_affine_parallelepiped` can be seen as generalizations of `draw_plane` and `draw_cube`, though some better interfaces with textures (uv coordinate, texture per face) and colors are probably still necessary.


not-fl3:
Thanks for PR!

PR #568, opened 2023-02-23
Add filtering mode for `Font` and `Atlas`

RybekU:
Simplest way to add filtering mode support for `Font`.
We could perhaps even go a step further and remove the `filter` field from the Atlas itself and just use the one on Texture2D.
That way we could bring back the commented out `Font::texture() -> Texture2D` instead and change filtering level (and more).

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -202,6 +202,25 @@ impl Font {
         }
     }
 
+    /// Sets the [FilterMode] of this font's texture atlas.
+    ///
+    /// Use Nearest if you need integer-ratio scaling for pixel art, for example.
+    ///
+    /// # Example
+    /// ```
+    /// # use macroquad::prelude::*;
+    /// # #[macroquad::main("test")]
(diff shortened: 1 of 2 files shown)


src/text.rs
```diff
@@ -202,6 +202,25 @@ impl Font {
...
+    /// # #[macroquad::main("test")]
+    /// # async fn main() {
+    /// let font = Font::default();
+    /// font.set_filter(FilterMode::Linear);
+    /// # }
+    /// ```
+    pub fn set_filter(&self, filter_mode: miniquad::FilterMode) {
+        let font = get_context().fonts_storage.get_font_mut(*self);
+
+        font.atlas.borrow_mut().set_filter(filter_mode);
```
not-fl3, line 221: Aren't all fonts share the same atlas? In other words, is it possible to set different filter modes to different fonts?

not-fl3:
Thanks for PR!

PR #570, opened 2023-02-27
Fix and update license name and formatting

GhtGhoster:
Cargo was updated and the usage of `license = "MIT/Apache-2.0"` was deprecated in favor of `license = "MIT OR Apache-2.0"`

Name of the MIT license in the first line of LICENSE-MIT suggesting it's an MIT/X Consortium license - which doesn't correspond to the license text, rust standard licensing or the license name in the Cargo.toml file - was changed to reflect that.


not-fl3:
Thanks!

PR #571, opened 2023-02-27
Adding mouse_delta_position function to Macroquad

poweron0102:
It is the fix for my last pull request.


not-fl3:
Thanks for PR!!

PR #572, opened 2023-03-01
Don't leak main future

matklad:
This helps with shutdown sequence, if, eg, the user-supplied futures holds onto some channels or some such.


matklad:
Also: https://matklad.github.io/2022/10/24/actions-permissions.html

Screenshot_20230301_225015

not-fl3:
> Also: https://matklad.github.io/2022/10/24/actions-permissions.html
> 
> Screenshot_20230301_225015

thanks! fixed :)

matklad:
Thanks! Didn't even have time for adding `std::process::exit(0)` work-around on my side :-)

PR #574, opened 2023-03-07
Added documentation for the Window2D struct, including a usage example

juli5n:
and some notes on the meaning of Camera2D fields.

### DIFF
--- a/src/camera.rs
+++ b/src/camera.rs
@@ -16,6 +16,27 @@ pub trait Camera {
 }
 
 #[derive(Clone, Copy)]
+/// # Example usage
+/// ```rust
+/// let camera = Camera2D {
+///     offset: Vec2::new(0.0,0.0), 
+///     zoom: Vec2::new(1.0,1.0),
+///     target: Vec2::new(1.0,1.0),
+///     ..Default::default()
+/// };
+/// 
+/// set_camera(&camera);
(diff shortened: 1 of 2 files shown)


juli5n:
After some more experimenting, I think that the two formulas in the documentation I wrote are wrong, but I cannot pinpoint the issue, sorry. Best regards, Julian!

juli5n:
Fedor, did you take a look at the documentation I wrote? I am not sure that it is entirely correct as I said above. I did the fmt and vec2 refactor as you asked and added some additional mouse-touch-simulation functionality. Best regards, Julian

juli5n:
I have not tested the touch-simulation mouse thing yet btw 🚨🚨!, I will get back to you once I have done that :)

src/input.rs
```diff
@@ -108,6 +112,41 @@ pub fn touches_local() -> Vec<Touch> {
         .collect()
 }
 
+fn update_mouse_touch_if_necessary() {
```
not-fl3, line 115: I feel like this doesn't belong to the documentation PR?

juli5n: Yes, that is not documentation, but rather a new feature, should I move it into another branch?

not-fl3: yes please

juli5n: Ok, I will try to do that later

juli5n:
(I tested it btw)

Issue #577, opened 2023-03-10
[Request] Missing set_mouse_cursor() method


viedevr:
I can't find the set_mouse_cursor() method. I am making my code editor, and want be able to change mouse cursor to 'Text' while it is over the editor. I found this in Miniquad's context. I think, it needs to be added to Macroquad.

PR #580, opened 2023-03-18
Add set_mouse_cursor

Eisverygoodletter:
This function was already in miniquad, but I saw https://github.com/not-fl3/macroquad/issues/577#issue-1619714957 and decided to try put it in macroquad.


not-fl3:
Hi! Thanks for PR! 

To be honest, as a user of macroquad, the game engine, I would expect `set_mouse_cursor` to set a cursor icon. It is a very rare use case when it is required to change a system cursor icon. And for this use cases - I think its okay to use raw miniquad's with `get_internal_gl`

PR #588, opened 2023-04-17
added draw_imglayer()

95ych:
#### Function  example:
Here an image layer (the orange cloud) is added with some offset and transparency in tiled as shown.
image
code:
```rust
#[macroquad::main("Typewriter")]
async fn main() {
	// load tiled-map.json, textures..etc
	...
	loop{
		...
		tiled_map.draw_tiles("platforms", dest_rect, None);
		tiled_map.draw_imglayer("bgimage", dest_rect, None);
		...
	}
	...
}
```
image

### DIFF
--- a/tiled/src/error.rs
+++ b/tiled/src/error.rs
@@ -11,6 +11,9 @@ pub enum Error {
     TextureNotFound {
         texture: String,
     },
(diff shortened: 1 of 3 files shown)


not-fl3:
Thanks for PR!

PR #589, opened 2023-04-18
Spawn only `amount` particles per cycle

RybekU:
Currently, the description of `EmitterConfig.amount` states `Amount of particles emitted in one emission cycle.`, however it only relies on comparison of `time_passed` to `EmitterConfig.lifetime`.
Due to this changing the `EmitterConfig.lifetime_randomness` configuration parameter to anything other than 0. will instead lead to incorrect amount of particles spawning when using the `EmitterConfig.one_shot` setting.
This pull request changes it so that the amount of particles emitted per one such cycle is always correct.

### DIFF
--- a/particles/src/lib.rs
+++ b/particles/src/lib.rs
@@ -488,6 +488,7 @@ pub struct Emitter {
 
     last_emit_time: f32,
     time_passed: f32,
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks!

PR #607, opened 2023-06-05
Improve safety of `get_quad_context`

e2-71828:
This uses a thread local variable to ensure that the `miniquad::Context` doesn't cross a thread boundary.  It also enforces the mutable reference aliasing rules by nulling the pointer when it is borrowed and restoring it when the borrow ends

This uncovered one instance of UB in `cargo test` due to an aliased mutable reference which has been fixed.  There may be others which aren't covered by the test suite; they will now panic.

This is a breaking change because the type of `get_internal_gl` has changed to include a guard object instead of an unbounded mutable reference; downstream users will need to adjust their code a bit (see the `raw_miniquad` and `particles` changes for an example of what this change may look like).

### DIFF
--- a/examples/raw_miniquad.rs
+++ b/examples/raw_miniquad.rs
@@ -4,8 +4,8 @@ use macroquad::prelude::*;
 async fn main() {
     let stage = {
         let InternalGlContext {
-            quad_context: ctx, ..
-        } = unsafe { get_internal_gl() };
+            quad_context: ref mut ctx, ..
+        } = &mut unsafe { get_internal_gl() };
 
         raw_miniquad::Stage::new(ctx)
     };
--- a/particles/src/lib.rs
+++ b/particles/src/lib.rs
@@ -506,8 +506,8 @@ impl Emitter {
 
     pub fn new(config: EmitterConfig) -> Emitter {
         let InternalGlContext {
-            quad_context: ctx, ..
-        } = unsafe { get_internal_gl() };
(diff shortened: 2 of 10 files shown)


e2-71828:
This is a possible model for fixing the soundness issues in #333.  I started with `get_quad_context` because it will prevent graphics calls from accidentally happening on the wrong thread, even if further changes don't happen.

e2-71828:
After asking about this on [URLO](https://users.rust-lang.org/t/unsafe-code-review-mutable-scoped-thread-locals/94885), it seems the current version isn't quite sound.  There's a couple of options to fix it:
1. Replace some of the panics with aborts, which will keep the API the same but make failures harder to debug
2. Replace `get_quad_context` with a callback-based API, which will eliminate most of the panics/aborts by requiring all of the accesses to be strictly nested.  This would also require `get_internal_gl` to change to something like `with_internal_gl<R>(impl FnOnce(InternalGl<'_>)->R)->R`, which will force a more complicated refacto on downstream users.

Is this general approach something you'd like me to work on further?  If so, do you have a preference for one of these two soundness fixes?

e2-71828:
`get_context()` returning `&'static mut` looks to still be a soundness problem in the new branch.  I suspect that a similar solution could work for that&mdash; In the new version, do you still assume all of the `macroquad` calls will be on a single thread, or are you aiming to make it thread safe somehow?

PR #608, opened 2023-06-06
Tweaks and bug fixes for snake.rs example

auberisky:
Hi!

I found two issues with the snake.rs example, which I addressed in this pull request:

- Game over text was not properly centered
- Implements navigation lock between two update cycles. Otherwise, the direction can be changed multiple times before the snake is updated, which can results in an illegal turn in the opposite direction.

Thanks,
Auberi


not-fl3:
Thanks for PR!

Issue #341, opened 2021-12-16
[Request] Add letterbox example


alexmozaidze:
Raylib has a letterbox example which enables to write games without worrying about different resolutions, it would be cool if Macroquad had something similar.

fwolfst:
Issue can be closed.

PR #612, opened 2023-07-02
A Letterbox example

LunarWarg:
A letterbox example like the raylib one. This should address #341


not-fl3:
Thats a great example, thanks for your work!

PR #616, opened 2023-07-14
Fix Android build

dm67x:
Replace the `get_quad_context` with `miniquad::window` to fix the build for android.


not-fl3:
If android somehow keeps calling editbox from another thread - its a serious problem that should be addressed properly. It is indeed a bug to try to access a single-threaded context from another thread and, unfortunately, the panic is legit.

PR #632, opened 2023-08-10
add an example for async file loading

lenscas:
it depends on https://github.com/not-fl3/miniquad/pull/405 which makes file loading also async on the desktop.

### DIFF
--- a/examples/async_file_loading.rs
+++ b/examples/async_file_loading.rs
@@ -1,0 +1,41 @@
+use macroquad::{
+    audio::{load_sound, play_sound, PlaySoundParams},
+    prelude::*,
+};
+
+#[macroquad::main("Audio")]
+async fn main() {
+    set_pc_assets_folder("examples");
+
+    let _coroutine = coroutines::start_coroutine(async move {
+        let sound_handle = load_sound("sound.wav").await;
+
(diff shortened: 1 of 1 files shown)


examples/async_file_loading.rs
```diff
@@ -0,0 +1,41 @@
...
+#[macroquad::main("Audio")]
+async fn main() {
+    set_pc_assets_folder("examples");
+
+    let _coroutine = coroutines::start_coroutine(async move {
+        let sound_handle = load_sound("sound.wav").await;
+
+        //to simulate it taking a while to load this file, lets do this in a nice loop :)
+
+        for v in 0..10000 {
```
not-fl3, line 15: I feel like this is a very very specific use case, that for sure deserver being documented, but I am not quite sure if `examples` is the best place for that. 

For `examples` I would rather spit those 10000 files into some reasonable chunks (like maybe 1000 per frame?)  and just load them through 10 frames, for the sake of code simplicity.

lenscas: Fair, I mostly made it for my own sake while testing https://github.com/not-fl3/miniquad/pull/405 and figured that I might as well push this one as well.

Issue #634, opened 2023-08-19
Make Image Safe


sloganking:
## Problem
Creating an image in `Macroquad` is verbose and unsafe. To create an image in macroquad I have to:
```rust
let mut image = Image::empty();
image.width = screen_width() as u16;
image.height = screen_height() as u16;
image.bytes = vec![0; image.width as usize * image.height as usize * 4];
```
- Create an image object
- Set it's width
- Set it's height
- Manually calculate the bytes necessary for it's dimensions and allocate them.
    - Look up the pixel format / size to determine needed bytes.

If you do ANY of this wrong. You'll get segfaults or unsafe and undefined behavior. 

## Feature request 
Why can't we have a safe image abstraction like the `Image` crate does? This is not only safer but also more convenient as it lets you create an image in one line.
```rust
let mut img = RgbImage::new(32, 32);
```

See [their docs](https://docs.rs/image/latest/image/struct.ImageBuffer.html#examples) for more.

PR #637, opened 2023-08-25
Added a safer constructor for Image

eliPerez12:
This PR is the solution to issue #634.

I've added a new constructor `Image::new(width, height) -> Image` , which abstracts away the process of making a new image, making it less error prone. Having to manually manage the amount of bytes needed was unnecessary and inefficient.

This is my first ever pull request so let me know if i missed anything or need to improve on anything!

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -85,6 +85,20 @@ impl Image {
         }
     }
 
+    /// Creates an empty Image from a width and height.
+    /// 
+    /// ```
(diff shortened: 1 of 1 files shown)


asmodehn:
Just jumping in here since I encountered issues with image manipulation for a small project:

We probably want to check overflows of usize type here. see for instance https://docs.rs/image/latest/src/image/buffer.rs.html#1271

Issue #618, opened 2023-07-15
Platformer example renders upside down


EriKWDev:
![image](https://github.com/not-fl3/macroquad/assets/19771356/7db4469c-b0df-4cd2-af89-c0ce1917bc49)
```yaml
macroquad 0.4.0 (git 8f2856f)
Linux fedora 6.3.12-200.fc38.x86_64 Thu Jul  6 04:05:18 UTC 2023 x86_64 GNU/Linux
cargo 1.70.0 (ec8a8a0ca 2023-04-25)
```
I'm using wayland

---

I checked the source code of `examples/platformer.rs` but didn't see any obvious culprit as to why this is. The tileset itself is rightside up.

Has the coordinate system recently changed?

Issue #578, opened 2023-03-11
`mouse_position_local` has y axis facing upwards, but in camera coordinates y axis faces downards


victoria-n-w:
The following code shows what I mean:

```
#[macroquad::main("Test")]
async fn main() {
    let camera = Camera2D::default();
    set_camera(&camera);

    loop {
        clear_background(BLACK);

        let mouse_pos = mouse_position_local();

        draw_circle(mouse_pos.x, mouse_pos.y, 0.01, RED);

        next_frame().await
    }
}
```

Expected: the red dot follows the mouse
Actual: the red dot mirrors the mouse's movement, flipped vertically

(unfortunately I was not able to record the screen, to show what I mean)

not-fl3:
Unfortunately this is a known issue - the `default` camera is flipped upside down :(
Fixing it will be too much of a breaking change, there are way too many projects with this `zoom: vec2(1, -1)` in their camera settings

IohannRabeson:
Hi! If this is just matter of removing one line of code, it's not so bad.

victoria-n-w:
I think it might confuse a lot of new people coming to macroquad - I spent 1 hour, trying to figure out why my code doesn't work. Maybe in this case, consistency is more important than backwards compatibility?

not-fl3:
I think we can release 0.4 with just a few changes, sub 1.0 all those breaking changes are totally fine

In other words, go ahead with a PR! :)

Issue #171, opened 2021-03-16
Camera2D.render_target is upside-down


brianwp3000:
The `render_target` field of a `Camera2D` is upside-down. I have included a repro--the gist is that if you draw a texture to a render_target and then draw that render_target to the screen the texture appears upside-down.

```rust
use macroquad::prelude::*;

#[macroquad::main("Texture")]
async fn main() {
    let mut render_target_camera = Camera2D::from_display_rect(Rect::new(0., 0., 800., 600.));
    render_target_camera.render_target = Some(render_target(800, 600));

    let ferris_texture: Texture2D = load_texture("examples/ferris.png").await;

    loop {
        // draw Ferris to the render_target texture
        set_camera(render_target_camera);

        clear_background(RED);
        draw_texture(
            ferris_texture,
            0.,
            0.,
            WHITE,
        );

        // draw the render_target texture to the screen
        set_camera(Camera2D::from_display_rect(Rect::new(0., 0., 800., 600.)));

        draw_texture(
            render_target_camera.render_target.unwrap().texture,
            0.,
            0.,
            WHITE,
        );

        next_frame().await
    }
}
```

![image](https://user-images.githubusercontent.com/9177969/111391703-64c5a400-8672-11eb-8653-9419a9d7d36c.png)

NicolasSabba:
I have the same issue in macroquad version 0.3.0-alpha.15.

For now I am using:

``` rust
draw_texture_ex(
      TEXTURE,
      0.0,
      0.0,
      WHITE,
      DrawTextureParams {
        flip_y: true,
        ..Default::default()
      },
    );
```

As a temporary solution works.

martin-t:
The behavior depends on whether there's a render target or not.

```rust
use macroquad::prelude::*;

#[macroquad::main("Camera bug")]
async fn main() {
    let render_target = render_target(600, 400);
    let rect = Rect::new(0.0, 0.0, 600.0, 400.0);
    let mut camera = Camera2D::from_display_rect(rect);
    //camera.zoom.y = -camera.zoom.y; // uncomment this to fix
    camera.render_target = Some(render_target); // or comment out this to fix

    loop {
        set_camera(&camera);

        clear_background(BLANK);
        draw_text("test", 50.0, 50.0, 32.0, GREEN);
        draw_line(0.0, 0.0, 300.0, 300.0, 1.0, GREEN);

        set_default_camera();
        draw_texture(render_target.texture, 0.0, 0.0, WHITE);

[...]

PR #638, opened 2023-08-26
Fix y inversion on examples

Mifom:
When building project with current macroquad version I'd found issue with camera. When a camera is set, the screen got inverted y.
After that I've found #618 and checked other examples. I've found that not only platformer has inverted y, but also tree and arkanoid. This examples use `Camera2D`  just like the platfornmer example. I've checked `Camera2D` code and found this inversion if not drawn on render target.

Am I right to remove inversion or this issues should be fixed other way?

### DIFF
--- a/src/camera.rs
+++ b/src/camera.rs
@@ -92,12 +92,7 @@ impl Camera for Camera2D {
         //   3. Move by -target
         let mat_origin = Mat4::from_translation(vec3(-self.target.x, -self.target.y, 0.0));
         let mat_rotation = Mat4::from_axis_angle(vec3(0.0, 0.0, 1.0), self.rotation.to_radians());
-        let invert_y = if self.render_target.is_some() {
-            1.0
-        } else {
-            -1.0
-        };
-        let mat_scale = Mat4::from_scale(vec3(self.zoom.x, self.zoom.y * invert_y, 1.0));
+        let mat_scale = Mat4::from_scale(vec3(self.zoom.x, self.zoom.y, 1.0));
         let mat_translation = Mat4::from_translation(vec3(self.offset.x, self.offset.y, 0.0));
 
         mat_translation * ((mat_scale * mat_rotation) * mat_origin)


not-fl3:
Hi, thanks for PR, camera's Y sure is messed up! 
Tho I am not quite sure if the proposed fix is really the solution: there are quite a few related problems  https://github.com/not-fl3/macroquad/issues/171

Mifom:
Okay, I'll look on related issues

Mifom:
I've done some research and found 3 issues related to this problem: #171, #578 and #618.

The second issue gave me thought about fixing problem. So now all this issues can be fixed using this PR.

Although, I think that drawing into camera has issue on miniquad side when `render_target` is not set (that's why the `if` before setting zoom camera is present).

BenMcAvoy:
So, I'm aware this was supposed to fix Y inversion but it used to work for me with zero issues. I'm now running the platformer example (copied into my own project with now changes) and it's inverted...

image

In the past when I ran this example (quite a while ago) things were fine. But I've noticed that it's upside down for me now.

BenMcAvoy:
NOTE: I'm on Fedora 38, BSPWM, X11, Nvidia drivers.

Mifom:
> In the past when I ran this example (quite a while ago) things were fine. But I've noticed that it's upside down for me now.

@BenMcAvoy  Thanks for the running. So as I understood, if you run `platformer` example on the master branch you'll have correct y rotation, and on the current branch - inverted, right?

Can you also try `letterbox` example? It's doing both drawing onto render target and on the screen, so would be valuable info

BenMcAvoy:
In the master branch, the `platformer` example is flipped. The `letterbox` example is not flipped.

image

Mifom:
> In the master branch, the `platformer` example is flipped. The `letterbox` example is not flipped.

Well, that's strange that with `master` and `fix_y_inversion` branches you have identical results as in new branch I've flipped results.

NOTE: I've noticed that you've said

> copied into my own project with now changes

... but there is no changes to `platformer` example, I've changed macroquad's camera logic.

BenMcAvoy:
Ah, sorry. I pulled changes and everything is working great! Platformer example isn't flipped and neither is letterbox on your branch (`fix_y_inversion`). Seems to work great, well done!

svenstaro:
Could this get merged? The current flip is quite confusing and this fix seems sound.

not-fl3:
This could be merged, but I am a little confused on how those axises should behave myself and where it is supposed to be up or down. 

I want to collect all the questionable cases somewhere (maybe here? https://github.com/not-fl3/quadtest) before merging this.

but I am super slow with that :|

PR #642, opened 2023-09-05
improving performance of getting the context and texture rendering

mlokr:
Hello,

When I was perf-ing my game I noticed function `same_thread` is eating significant amount of my cpu along with `draw_texture_ex`. I modified the code to perform better, for `same_thread`, thread local is used to cache it. I do not know how good this performs on all platforms but on linux it makes the check almost free. For the texture render, I noticed calls to width and height are internally calling methods on trait object so rust cannot know if its safe to reuse the values in other calls, I also eliminated texture clone which removed atomic operations.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,9 +168,11 @@ pub(crate) mod thread_assert {
 
     pub fn same_thread() {
         unsafe {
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks for PR!

PR #647, opened 2023-09-15
Added a "serde" feature.

jackolantern:
This commit:
- adds a "serde" feature.
- adds the serde crate as an optional dependency, activated by enabling the serde feature.
- enables the serde feature of the `glam` dependency when the new serde feature is activated
- annotates the `Color` structure with serde's Serialize and Deserialize traits

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ readme="README.md"
 [features]
 audio = ["quad-snd"]
 log-rs = ["log"]
+serde = ["dep:serde", "glam/serde"]
 default = []
 
 [package.metadata.android]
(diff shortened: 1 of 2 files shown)


not-fl3:
Cargo features are additive, so you can always enable glam/serde from the top-level crate

For the `serde` for Color - I do not think depending on the whole serde is worth it for just one struct.

JMANN240:
Any thoughts on reopening this? I know #683 added serde support for glam types, behind the optional `glam-serde` feature, but it would be nice to also have an optional `color-serde` feature (or even just an optional `serde` feature for both.

PR #648, opened 2023-09-17
Added Num blanket implementation

tynberry:
Adds blanket implementation for Num so that more types can be included.

Right now there are only u32 and f32, which is lacking. (for example one might want to use f64)
This fixes it and so long a type implements the sub traits of Num the Num will be automatically added, allowing for user types to be also used in Drag with minimal friction.

(also it removes the impls for u32 and f32, since they are not needed)


tynberry:
Okay, the rabbit hole goes deeper than expected. Closing it for now...

PR #650, opened 2023-09-18
Wrap fontdue::Font into Arc

ozkriff:
Wrap fontdue::Font into Arc to make ::Font relatively cheap to clone again.


not-fl3:
Welcome back @ozkriff \0/

PR #654, opened 2023-09-30
Added AnimatedSprite::is_last_frame

FourteenBrush:
Added this method to implement a state transitioning system, in where a sprite can only change its animation when it's fully ended.


not-fl3:
Thanks for PR!

PR #656, opened 2023-10-06
Seperate texture from the mesh object

ryanpeach:
I personally need a set of vertices and indices that update very irregularly (never) but a texture that updates every frame. The vertices and indices copying every frame is not ideal, and I haven't seen many libraries coupling them like this. This proposal is a lot cleaner in my codebase at the very least, because I can use a borrow for a permanent "mesh" object and a borrow for a new texture object. No copying the texture into the mesh, which prevents me from needing to re-make the mesh every frame, which prevents me from needing to do a clone on my vertices and indices every frame.

Profiling wise in my application it wasn't such a big deal, but every little clone does matter.

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -25,13 +25,12 @@ impl From<Vertex> for crate::quad_gl::VertexInterop {
 pub struct Mesh {
     pub vertices: Vec<Vertex>,
     pub indices: Vec<u16>,
-    pub texture: Option<Texture2D>,
 }
 
-pub fn draw_mesh(mesh: &Mesh) {
+pub fn draw_mesh(mesh: &Mesh, texture: Option<&Texture2D>) {
     let context = get_context();
 
-    context.gl.texture(mesh.texture.as_ref());
+    context.gl.texture(texture);
     context.gl.draw_mode(DrawMode::Triangles);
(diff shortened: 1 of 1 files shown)


ryanpeach:
I found a way to get around this using borrowing. Something like this works:

```rust

    // Pre-compute all vertices and indices
    let mut all_indices: Vec<Vec<u16>> = radial_mesh.get_indices();
    let mut all_vertices: Vec<Vec<Vertex>> = radial_mesh.get_vertices();

    loop {
        ...
        
        // Generate new textures and draw them
        let all_textures = radial_mesh.get_textures();
        for (i, texture) in all_textures.into_iter().enumerate() {
            let vertices = replace(&mut all_vertices[i], Vec::new());
            let indices = replace(&mut all_indices[i], Vec::new());
            let mesh = Mesh{
                vertices,
                indices,
                texture: Some(texture),
            };
            draw_mesh(&mesh);
            let _ = replace(&mut all_vertices[i], mesh.vertices);
            let _ = replace(&mut all_indices[i], mesh.indices);
        }
        
    }
```

It's a lot more verbose, but it will do.

not-fl3:
I think this change makes sense, mind if I'll reopen and merge it?

PR #660, opened 2023-10-18
colors: add alpha() function to modify constant colors easily

naufraghi:
Like `RED.alpha(0.5)`

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -105,6 +105,10 @@ impl Color {
     pub fn from_vec(vec: glam::Vec4) -> Self {
         Self::new(vec.x, vec.y, vec.z, vec.w)
     }
+
+    pub fn alpha(&self, a: f32) -> Self {
+        Self::new(self.r, self.g, self.b, a)
+    }
 }
 
 pub mod colors {


naufraghi:
Just a suggestion, was handy in my exploration. Feel free to close the MR as you wish. Kudos for macroquad!

not-fl3:
I like how `RED.alpha(0.5)` looks, and it is indeed shorter and cleaner than `Color {a: 0.5, ..RED}`
But I do not really like that other color components doesn't work like this and neither do vectors/any other structs we got.

To be honest I do not have a strong enough opinion on consistency vs convinience so I can't make a decision on merging it or not.

naufraghi:
My ratio was:

1. `a` is a bit special as a color component, the color is the same, but with a special transparency feature. Changing `rgb` makes a new color,
2. Constant colors are a limited number, and usually used arbitrarily just to put something on the screen. In this setting it's handy for exploration to see below something, hence the `alpha` modifier.

But, I completely overlooked the `Color {a: 0.5, ..RED}` option, it works well enough without any unbalance in the API!

Issue #655, opened 2023-10-03
`get_screen_data` leaks memory and is not deallocated when the program exits


martin-t:
Using macroquad 0.4.4:

```rust
use macroquad::prelude::*;

#[macroquad::main("BasicShapes")]
async fn main() {
    loop {
        get_screen_data();

        next_frame().await
    }
}
```

This will make memory usage grow very fast until it crashes the system.

Memory usage does not decrease by quitting the program, i had to use `sudo zsh -c 'echo 3 >| /proc/sys/vm/drop_caches'`. Using Kubuntu 22.04.

Afrovis:
Same issue on Mac when using the get_screen_data() function in the loop. Memory does decrease on crash/quitting though.

ElnuDev:
This issue should have been closed by #666, no?

PR #666, opened 2023-11-05
Fixed memory leak at get_screen_data()

Ordasto:
Changed get_screen_data() to use a managed macroquad texture with `store_texture` instead of `from_miniquad_texture` to fix the memory leak mentioned in #655.

I'm not sure if there was a specific reason it was using `from_miniquad_texture` before but I didn't notice any difference in behaviour.


not-fl3:
Thanks for PR!

Issue #575, opened 2023-03-07
[Feature Request] Ellipses


alexmozaidze:
Circles are cool and all, but aren't ellipses just plain cooler?

SmartManLudo:
I made something 
![image](https://github.com/not-fl3/macroquad/assets/92421677/30d7ee3d-de2c-4a89-ace7-3fdc1054a25a)
Currently only works to draw lines. If you wanted to draw a filled ellipse then set the thickness number to something big (like 100, 1000 just filles the entire screen) however it takes more space than specified.
Heres the same ellipse but with more thickness:
![image](https://github.com/not-fl3/macroquad/assets/92421677/e0a3c36e-8a1b-408c-817e-5dc0f6ded1cb)

Here's the code:
```rs
fn draw_ellipse_line(center_x:f32,center_y:f32,rx:f32,ry:f32,num_segments:u32,thickness:f32,color:Color) {

        for i in 0..num_segments {
            let angle = 2.0 * std::f32::consts::PI * (i as f32) / (num_segments as f32);
            let x = center_x + rx * angle.cos();
            let y = center_y + ry * angle.sin();

            draw_circle(x,y,thickness,color);
        }
}
```
Example usage:
```rs
#[macroquad::main("Draw Ellipse")]
async fn main() {
    const WIDTH: f32 = 800.0;
    const HEIGHT: f32 = 600.0;
    const CENTER_X: f32 = WIDTH / 2.0;
    const CENTER_Y: f32 = HEIGHT / 2.0;
    const RX: f32 = 100.0;
    const RY: f32 = 200.0;
    const NUM_SEGMENTS: u32 = 1000;

    loop {
        clear_background(BLACK);

        draw_ellipse_line(CENTER_X,CENTER_Y,RX,RY,NUM_SEGMENTS,1.0,RED);
        next_frame().await;
    }
}
```
Code was referred from [OpenGl/ellipse.cpp](https://github.com/huzaifafaruqui/OpenGL/blob/master/ellipse.cpp)

PR #668, opened 2023-11-11
draw_ellipse and draw_ellipse_lines

Murrent:
Solid ellipses and lined ellipses as proposed in #575 

This implementation draws a polygon with 20 sides just like how circles are drawn, but this one can be stretched and rotated.

image


not-fl3:
Ohh, I totally missed #575

Thanks for PR!

PR #671, opened 2023-11-29
Add draw_grid_ex, and have draw_grid use that with vec3(0,0,0)

computermouth:
It always bugged me that you couldn't move the grid in raylib. I'd considered also adding the ability to do like an XY grid, XZ grid, YZ grid, but couldn't think of an intelligent way to do it.

Maybe setting an `up: Vec3`?

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -3,7 +3,7 @@
 use crate::{color::Color, get_context};
 
 use crate::{quad_gl::DrawMode, texture::Texture2D};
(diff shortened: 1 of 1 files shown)


computermouth:
Updated to now allow rotation with a Quat

2023-11-30--1701372858_772x683_scrot

the code for ^this is here: https://github.com/computermouth/macroquad-test/commit/3bc38c6f5b600aea4f02c68106abfcf5c371b5f8

not-fl3:
Thanks for PR!

Issue #649, opened 2023-09-17
Command-Q does not quit program on macOS


alterae:
On macOS, the customary shortcut to quit any application is Command-Q (LeftSuper-Q using macroquad KeyCodes). However, this does not work in macroquad applications.

While it is trivial to implement by hand, it is still a bit annoying and mac users generally expect Command-Q to work on all platforms.

If this could be implemented in the engine, that would be excellent.

markmurphydev:
I've submitted not-fl3/miniquad#464 to add a menu bar entry with a "quit application" shortcut. This is the normal way to implement the command-q shortcut on osx, and lets you rebind it in system keyboard settings.

Note that on osx on master, clicking on a menu bar item currently freezes the window. This is due to not-fl3/miniquad#455.

The Winit crate also adds by default the "About application", "Services", "Hide application", "Hide others", and "Show all" menu items. These are also standard Mac menu items, with system-default implementations. I can add them if we want.

PR #674, opened 2023-12-09
Impl cmd q macos

Num0Programmer:
Adds Cmd+Q functionality for MacOS - issue #649.

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -634,7 +634,7 @@ impl InputHandler for Ui {
         if key == KeyCode::Enter {
             self.input.enter = true;
         }
-        if key == KeyCode::Q {
+        if cmd && key == KeyCode::Q {
             self.input.quit = true;
         }
 
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -70,7 +70,7 @@ use crate::{
         Font,
     },
     texture::Image,
-    ui::{canvas::DrawCanvas, render::Painter},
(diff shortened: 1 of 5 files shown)


src/ui.rs
```diff
@@ -1275,7 +1283,10 @@ pub(crate) mod ui_context {
                 || is_key_pressed(KeyCode::LeftControl)
                 || is_key_pressed(KeyCode::RightControl)
             {
-                ui.key_down(megaui::KeyCode::Control, shift, ctrl);
+                ui.key_down(megaui::KeyCode::Control, shift, ctrl, cmd);
+            }
+            if ui.input.quit {
+                get_context().quit_requested = true;
```
not-fl3, line 1289: macroquad::ui is a module about drawing game UI, I am not quite sure if handling platform quit request belongs here?

Num0Programmer: I am inclined to agree. Would this be better moved to something like begin_frame() in src/lib.rs, or could you make a suggestion where a platform quit request would be better handled?

not-fl3: begin_frame looks like a right place to me. 

Justa side question, but is manual handling Cmd-Q a common practice on MacOs? Isnt there some sort of a system-level hotkeys for this kind of things?

Num0Programmer: I only thought about that after I last commented. I can do some research; I bet I can find something. I will get back to this in a day or two with relevant changes.

PR #675, opened 2023-12-21
Consider window margin when setting position

jwarlander:
When creating a window with `movable` = `false`, the layout of components is a bit off. This is caused by the call to `set_position()` here:

https://github.com/not-fl3/macroquad/blob/d8e2f9d2128d9bb14a22f225efebfbf41f5ec979/src/ui.rs#L795-L797

..which ends  up setting the position itself as expected, but failing to adjust for window margins for the cursor area.


not-fl3:
Thanks for PR!

Issue #667, opened 2023-11-08
Arkanoid is mirrored and upside down


HazeSystem:
No idea what the problem is, but yeah. Running on Windows 10.

![2023-11-08 14_29_40-Arkanoid](https://github.com/not-fl3/macroquad/assets/37942530/5beac67e-7c19-44fc-ada0-757e566c0107)

PR #678, opened 2023-12-27
Display `arkanoid` and `platformer` the right way up (#667)

pbevin:
Commit 3887953 inverted the Y-axis; this fixes the two examples where it matters.


not-fl3:
Thanks for PR!

PR #683, opened 2024-01-08
Support of glam types serialization

ihor-tarasov:


not-fl3:
Note that cargo's feature are additive - you can add glam/serde to your own project and all the dependencies in the depenency tree will get serde feature!

iiYese:
> Note that cargo's feature are additive - you can add glam/serde to your own project and all the dependencies in the depenency tree will get serde feature!

This relies on unification which sucks because you have to match the glam versions. The latest release of macroquad is like 8 versions behind the latest glam version & the main branch is 1 version behind. The only place you can get that information the dep tree.

not-fl3:
If we merge it, we force every single macroquad user to depend on serde and compile serde, syn, proc_macro and whatnot all the time. 

I we do not merge it, each user have a choice to enable glam's serialization and which serialization library to use with glam if they choose so.

Considering macroquad's priorities, here I choose slightly less usability for serde users over slower compile time for everyone else.

iiYese:
> If we merge it, we force every single macroquad user to depend on serde and compile serde, syn, proc_macro and whatnot all the time.
> 
> I we do not merge it, each user have a choice to enable glam's serialization and which serialization library to use with glam if they choose so.
> 
> Considering macroquad's priorities, here I choose slightly less usability for serde users over slower compile time for everyone else.

Where did you get this notion? The whole point of feature flags is conditional compilation. Unless you set it as a default feature which this change does not do there is no effect on compile times. You are also not just choosing "slightly less usability for serde users" you are choosing slightly less usability for a significant amount of current and potential macroquad users because those venn diagrams intersect.

not-fl3:
Ohhh, I missread the PR :facepalm: it really is not a deafault feature, I agree, there is no harm in having an extra feature. Sorry!

PR #684, opened 2024-01-10
Add a function to load an OBJ mesh with texture

0x177:
a comment on the top of model.rs mentions loading models, but there isn't any code that does that, so i added it. this function is probably slow and i don't like the tobj dependency, but it is easy and less bloated than other obj parsers.

Blender monkey with brick texture applied
Blender monkey with brick texture applied


not-fl3:
Hi, thanks for PR! 

Tho I am not quite sure if this code justify adding an extra dependency - user can always use `tobj` directly, its not that much code to get `tobj` data into macroquad's model. 
And, lets be honest, macroquad is not really very good for 3d anyway, so making every single user to compile `tobj` all the time might be a bit of an overkill here.

PR #687, opened 2024-01-15
fix texture size bug when using batched_texture

jwvg0425:
Fixed a bug in the draw_texture_ex function where the width and height values were not updated to the width and height of the batched_texture, causing the values of the vertices to become incorrect.


not-fl3:
Thanks!

PR #688, opened 2024-01-15
Change to break_batching update only when pipeline's are changed

jwvg0425:
When setting the material using gl_use_material, if the same material is passed, there is no need to break batching and increase draw calls. It seems that not performing break_batching when the pipeline is the same would be better in terms of performance.


not-fl3:
Thanks for PR!

Issue #664, opened 2023-10-24
Initialize button with texture doesn't work


WindowGenerator:
I tried initialize button with png texture from example [ui](https://github.com/not-fl3/macroquad/blob/master/examples/ui.rs), but this [place](https://github.com/not-fl3/macroquad/blob/master/examples/ui.rs#L209) in the example doesn't work. Below I compared last worked version with latest.

macroquad=0.3.26
![image](https://github.com/not-fl3/macroquad/assets/41333030/f9064e25-d1d5-49f0-9acf-dea8b3b13f30)

macroquad=0.4.4
![image](https://github.com/not-fl3/macroquad/assets/41333030/574caf9f-c2c3-4476-bcb5-ace4f3fa7121)

OS:
```
NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
```

streak324:
https://github.com/not-fl3/macroquad/blob/8f2856f09674324b1222d7f344b2d101377430bb/src/ui/render/mesh_rasterizer.rs#L250-L253

removing this if check gets the textures showing again. however I'm not sure what unintended behavior would result from the check being removed.

PR #696, opened 2024-02-03
fix bug with checking last drawlist was texture

streak324:
This change addresses a bug that appears to prevent new drawlists from being pushed if the most recent drawlist has no texture. 
I believe the intention of the code is to only prevent new drawlists if the most recent drawlist does have a texture and its the same as the command

This closes https://github.com/not-fl3/macroquad/issues/664


not-fl3:
Nice, thanks for PR!

PR #699, opened 2024-02-11
fixed a bug in draw_rectangle_ex()

AdamSlayer:
Two vertices had the same uv coordinate, which makes no sense. It now has the same uv coordinates as draw_rectangle()


not-fl3:
Thanks for PR!

PR #703, opened 2024-03-02
fix: objectlayer -> objectgroup for layer case

thatguydoru:
As per tiled's [Layer](https://doc.mapeditor.org/en/stable/reference/json-map-format/#layer) JSON doc.


not-fl3:
Thanks for PR!

PR #705, opened 2024-03-12
Configurable clear_background() each frame

DuckyBlender:
Hi guys
This is my first pull request so I wanted to do something easier to do, but I still failed. Any ideas what I'm doing wrong? When setting `clear_background` in `window_conf()` to `false` it still clears the background...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -378,7 +378,7 @@ impl Context {
         self.ui_context.process_input();
 
         // Check if the window conf "clear_background" is set to false
-        if miniquad::window::get_clear_background() {
+        if miniquad::window::clear_background() {
             let color = Self::DEFAULT_BG_COLOR;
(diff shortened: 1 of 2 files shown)


DuckyBlender:
Oh I just noticed I also need to modify the `miniquad` crate and my changes didn't include in the commit... Should I open a pull request there too?

DuckyBlender:
Opened https://github.com/not-fl3/miniquad/pull/421

not-fl3:
Not clearing the background is not really possible - imagine what would happen if the window got resized between two frames? Also drivers do not guarantee that window content should be persisted if nothing is being drawn between the frames on a lot of platforms. 

Correct macroquad-level solution would be in always rendering into a texture, drawing this texture on each frame, and keep clear_background optional this way. But I am not sure if this should be done inside macroquad or left for user to decide/implement.

PR #711, opened 2024-04-10
Added update_from_bytes method to Texture2D

Nrosa01:
This allows user to update a texture from an array u8. If you have a buffer from some crate now you can update a texture without depending on macroquad image.

This change affects me directly and thus I implemented it. I hope this is compatible with macroquad way of doing things.


not-fl3:
Thanks for PR!

PR #713, opened 2024-04-11
Allowed users to optinoally define number of sides on draw_circle and draw_circle_lines

Nrosa01:
Circles look really bad with only 20 sides. Optionally being able to change that number will be good. I wanted to avoid creating a new method just for this so this is a breaking change as Rust doesn't support default field values for function parameters user will be forced to either add Some(u8) or None to these two function calls in they codebase. They can easily fix it with regex but it's still a little annoying breaking change for something simple. Maybe this can be merged when a new macroquad version releases.

### DIFF
--- a/src/shapes.rs
+++ b/src/shapes.rs
@@ -204,14 +204,15 @@ pub fn draw_poly_lines(
 }
 
 /// Draws a solid circle centered at `[x, y]` with a given radius `r` and `color`.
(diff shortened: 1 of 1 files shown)


cyrgani:
You should at least update all usages of this fuunction inside of macroquad so CI no longer fails and the crate can be built.

Issue #693, opened 2024-01-26
coroutine_pause.rs test broken on head


mandroll:
When running `cargo test` in my fork, I see the following error:

```sh
cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.07s
     Running unittests src\lib.rs (target\debug\deps\macroquad-39e857bc1218d84e.exe)

running 2 tests
test color::color_from_bytes ... ok
test material::shaders::preprocessor_test ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests\back_to_the_future.rs (target\debug\deps\back_to_the_future-29cb4bf128cc5937.exe)

running 1 test
test back_to_the_future ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.28s

     Running tests\back_to_the_future_coroutines.rs (target\debug\deps\back_to_the_future_coroutines-b2e3d9943e110d30.exe)

running 1 test
test back_to_the_future_coroutine ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.41s

     Running tests\coroutine_pause.rs (target\debug\deps\coroutine_pause-8289b31e25dc5fee.exe)

running 3 tests
test coroutine_execution_order ... ok
error: test failed, to rerun pass `--test coroutine_pause`

Caused by:
  process didn't exit successfully: `C:\Users\mandroll\Documents\GitHub\macroquad\target\debug\deps\coroutine_pause-8289b31e25dc5fee.exe` (exit code: 0xc0000374, STATUS_HEAP_CORRUPTION)
note: test exited abnormally; to see the full output pass --nocapture to the harness.
```

cyrgani:
The error comes from this `Drop` implementation:
```rust
impl Drop for GlPipelineGuarded {
    fn drop(&mut self) {
        get_context().gl.delete_pipeline(self.0);
    }
}
```

The mutable static `CONTEXT` contains a `Option<Context>` object. `get_context` returns a `&'static mut Context`.

A `Context` contains a `UiContext`, which contains a `Option<Material>`. A `Material` finally contains an `Arc<GlPipelineGuarded>`.

Since `CONTEXT` is a static variable, its value is not dropped at the end of the program.

[...]

PR #715, opened 2024-04-13
Implement Clone, Copy and Debug on possible structs and enums

musjj:
I was really annoyed with getting blocked by some structs not implementing `Clone`/`Copy`/`Debug` so I tried to implement them in as many places as possible (sorry for the fat diff).

### DIFF
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -6,7 +6,7 @@ use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
 use crate::Error;
 
 // Returns Pending as long as its inner bool is false.
-#[derive(Default)]
+#[derive(Default, Clone, Copy, Debug)]
 pub struct FrameFuture {
     done: bool,
 }
@@ -27,6 +27,7 @@ impl Future for FrameFuture {
     }
 }
 
+#[derive(Clone, Debug)]
 pub struct FileLoadingFuture {
     pub contents: Arc<Mutex<Option<Result<Vec<u8>, Error>>>>,
 }
--- a/src/telemetry.rs
+++ b/src/telemetry.rs
@@ -47,6 +47,7 @@ impl Zone {
     }
 }
 
+#[derive(Clone, Debug)]
 pub struct ZoneGuard {
     _marker: (),
 }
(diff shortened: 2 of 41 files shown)


src/exec.rs
```diff
@@ -6,7 +6,7 @@ use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
 use crate::Error;
 
 // Returns Pending as long as its inner bool is false.
-#[derive(Default)]
+#[derive(Default, Clone, Copy, Debug)]
```
not-fl3, line 9: I did not verified so I don't really know, but is it really safe to do that for macroquad's futures? What implications does it bring?

src/telemetry.rs
```diff
@@ -47,6 +47,7 @@ impl Zone {
     }
 }
 
+#[derive(Clone, Debug)]
```
not-fl3, line 50: I am not quite sure if cloning guard struct is useful?

src/exec.rs
musjj: Yeah, I think I might've been too aggressive with the impls. Let me know if there's anything else unsound.

src/telemetry.rs
musjj: Reverted this one too.

musjj:
Btw, I'm unable to run the tests on my machine for some reason. Even with a fresh unmodified clone I'm getting errors such as:
```rust
test coroutine_execution_order ... ok
double free or corruption (out)
```
The errors seems to be inconsistent between runs, so I'm guessing that there's a race condition somewhere.

Could you run it on your side if possible?

profan:
> Btw, I'm unable to run the tests on my machine for some reason. Even with a fresh unmodified clone of `master`, I'm getting errors such as:
> 
> ```rust
> test coroutine_execution_order ... ok
> double free or corruption (out)
> error: test failed, to rerun pass `--test coroutine_pause`
> ```
> 
> The errors seems to be inconsistent between runs, so I'm guessing that there's a race condition somewhere.
> 
> Could you run it on your side if possible?

Random but this specific test error seems to be exploding for me as well, likely everyone right now

mandroll:
>>Btw, I'm unable to run the tests on my machine for some reason. Even with a fresh unmodified clone of master, I'm getting errors ...

>Random aside here but this specific test seems to be exploding for me as well, likely everyone right now

Documented here: #693

PR #716, opened 2024-04-17
replace std::f32::{EPSILON, MAX} with f32:: as per rust docs recommendation (old way is deprecated)

narodnik:
replace std::f32::{EPSILON, MAX} with f32:: as per rust docs recommendation (old way is deprecated)


not-fl3:
Thanks for PR!

Issue #635, opened 2023-08-19
Texture2D memory leak


sloganking:
`Texture2D` types are never removed from memory. 

Memory steadily grows until system crash when repeatedly converting an `Image` into a `Texture2D` in a loop. Even though the `Texture2D` is dropped at the end of each loop.

![image](https://github.com/not-fl3/macroquad/assets/16965931/ff2c631f-08dc-496e-870c-79b5ad8c79b0)

Memory does not grow when the `macroquad` program is closed or my `Image`s are never converted to `Texture2D`s
![image](https://github.com/not-fl3/macroquad/assets/16965931/e8d38fbc-bebe-4229-9e6e-30c75bbc2964)

Afrovis:
I'm having an issue that I believe is related when calling get_screen_data() in my render loop (for image saving per frame). Not fixed in 0.4.4 yet.

eboatwright:
This is a problem in my project as well, I'm making a pixel art game, but I still want the window resizable, so every frame I'm creating a render_target and a Camera2D every frame, with the current resolution and after the program runs for a bit my CPU usage spikes, and my computer crashes

mandroll:
I also have this issue. Here is a minimal reproducer:

```rs
use macroquad::prelude::*;

#[macroquad::main("Texture test")]
async fn main() {
    let texture_size: u32= 64 * 16;
    loop {
        let render_target = render_target(texture_size, texture_size);
        next_frame().await;
    }
}
```

Another user on Mac claims the issue does not happen for them, suggesting it may be platform-specific. In my case, I am on Windows 10.

mandroll:
I can confirm the bug is still present on HEAD. I have tested with:

```toml
[package]
name = "texture-memory"
version = "0.1.0"
edition = "2021"

[dependencies]
macroquad = { git = "https://github.com/not-fl3/macroquad.git", branch = "master" }
```

mandroll:
Here is a [dhat](https://docs.rs/dhat/latest/dhat/) heap profile of the above program: https://gist.github.com/mandroll/02b19220221bc81a1f83d2d299914a32

ElnuDev:
Any news on this? This is a very serious issue. I'm pretty sure this is the same 

> I also have this issue. Here is a minimal reproducer:
> 
> ```rust
> use macroquad::prelude::*;
> 
> #[macroquad::main("Texture test")]
> async fn main() {
>     let texture_size: u32= 64 * 16;
>     loop {
[...]

Issue #628, opened 2023-08-06
Sound memory leak


andreymal:
```rust
loop {
    load_sound("./sound.ogg").await?;
}
```

RAM is leaking.

Like with font atlas, I tried to implement `Drop` for `Sound`:

```rust
impl Drop for Sound {
    fn drop(&mut self) {
        if Arc::strong_count(&self.0) < 2 {
            self.0.delete(&get_context().audio_context.native_ctx);
        }
    }
}
```

and this fixed the leak (but I'm not sure if this solution is complete and/or correct because I don't even know why `Arc` is used here)

not-fl3:
Hopefully fixed with #813

PR #719, opened 2024-04-30
Fix issues with RenderTarget leaking memory.

profan:
* Fix issue with RenderTarget leaking memory by adding a drop impl, reported in issue #635

The memory leak issues just floating around were bothering me, and seemed simple enough to fix so figured why not :)

I also looked at #628 but couldn't actually reproduce the issue, so left that one alone for now.

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -230,24 +230,27 @@ impl Image {
             bytes,
         }
     }
-    
+
     /// Blends this image with another image (of identical dimensions)
     /// Inspired by  OpenCV saturated blending
     pub fn blend(&mut self, other: &Image) {
-        assert!(self.width as usize * self.height as usize == other.width as usize * other.height as usize);
+        assert!(
+            self.width as usize * self.height as usize
+                == other.width as usize * other.height as usize
+        );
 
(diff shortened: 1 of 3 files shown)


src/texture.rs
```diff
@@ -340,6 +346,12 @@ pub struct RenderTarget {
     pub render_pass: miniquad::RenderPass,
 }
 
+impl Drop for RenderTarget {
```
not-fl3, line 349: will this lead to a use after free kind of error? 

```
{
  let a = render_target.clone(); 
}
draw_texture(render_target);
```

profan: right, yeah assuming clone only copies the underlying resources in a shallow way, i guess it would yeah

it seems like both the Texture2D and RenderPass held are basically unmanaged, so you'd have to explicitly do the copy presumably in the Clone impl for RenderTarget

profan:
Let me split this PR into two parts so that the Atlas leak issue can be fixed separately of this bigger RenderTarget/RenderPass issue quickly

not-fl3:
Thanks for PR!

Issue #627, opened 2023-08-06
Font atlas memory leak


andreymal:
```rust
loop {
    load_ttf_font("./font.ttf").await?;
}
```

Both system RAM and GPU memory are leaking.

I tried to implement `Drop` for `Atlas` like in https://github.com/not-fl3/macroquad/commit/65451f5bdcf7802c3d62181873d30b9c3c6c49f2:

```rust
impl Drop for Atlas {
    fn drop(&mut self) {
        get_context().quad_context.delete_texture(self.texture);
    }
}
```

and this fixed the leak (but I'm not sure if this solution is complete and/or correct)

PR #721, opened 2024-05-02
Fix issues with Atlas leaking memory.

profan:
* Fix issue with Font Atlas leaking memory by adding a drop impl, reported in issue #627

Split this out from the other PR so at least this change would go in, since it seems mostly unproblematic :)

### DIFF
--- a/src/text/atlas.rs
+++ b/src/text/atlas.rs
@@ -34,7 +34,8 @@ pub struct Atlas {
 
 impl Drop for Atlas {
     fn drop(&mut self) {
-        self.delete();
+        let ctx = &mut get_context().quad_context;
+        ctx.delete_texture(self.texture);
     }
 }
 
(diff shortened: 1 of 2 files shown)


profan:
I guess the only question here might be, maybe that delete function does not need to be public?

src/text/atlas.rs
```diff
@@ -184,4 +190,8 @@ impl Atlas {
             );
         }
     }
+    fn delete(&mut self) {
+        let ctx = &mut get_context().quad_context;
+        ctx.delete_texture(self.texture);
```
not-fl3, line 195: I believe we used to have `delete` on everything, but now, with Drop, should it really be a separate function, not just `drop`?

profan: Yeah that's fair, it's probably an entirely unnecessary convention to follow now, I'll remove the delete function and just have the Drop impl :)

not-fl3:
Thanks for PR! :+1:

PR #726, opened 2024-05-27
glam => 0.27

jack-cooper:
Updates `glam` to version `0.27`.

My main motivation here is to at least hit `0.24` for the convenience of the `saturating_*` functions on the integer vector types.

I get a few unrelated test failures locally so I can't 100% verify that this is ok, but output is at least identical to running tests on  `master` which is promising. For completeness, the failures are the coroutine tests as noted at #693, and a number of doc tests. Running anything locally with this build hasn't caused me any problems on either MacOS or Linux X11.


not-fl3:
Thanks! Pretty sure the issues are not glam related!

PR #728, opened 2024-05-29
remove debugging println! call

cyrgani:
A `println!` call in the middle of a function was added in 246cbfc and made it into version 0.4.7. This appears to be a debugging leftover that should be removed. This PR removes it.


not-fl3:
ohh my bad

PR #733, opened 2024-06-13
Implementation of functions to draw cylinders and cones

jonmue42:
Adds functions to draw cylinders.
Can also be used for cones by setting on of the top or bottom radius to zero.

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -583,7 +583,6 @@ impl Default for DrawCylinderParams {
             draw_mode: DrawMode::Triangles,
         }
     }
-    
 }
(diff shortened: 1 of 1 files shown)


not-fl3:
LGTM! Make the CI happy again and I'll merge right away!

jonmue42:
Alright I ran cargo fmt for the changed models.rs file, so it should pass now.
Note: The formatter also complains about something in the shapes.rs file, but I obviously didn't touch that one.

not-fl3:
Perfect, thanks!

PR #736, opened 2024-06-16
Update shapes.rs, add draw_arc

KurlykovDanila:
Add draw_arc function


KurlykovDanila:
Add draw_arc function with extended and corrected behavior of draw_poly_lines

not-fl3:
Thanks for PR!

PR #738, opened 2024-06-16
Changes draw_text_ex to pass params by reference

maxpiepenbrink:
If you're re-using `TextParams` for various things, it can be a little clumsy to need to `.clone()` it everywhere.

### DIFF
--- a/examples/arkanoid.rs
+++ b/examples/arkanoid.rs
@@ -53,7 +53,7 @@ async fn main() {
                 "Press space to start",
                 SCR_W / 2. - 5.,
                 SCR_H / 2.,
-                text_params,
+                &text_params,
             );
 
             ball_x = platform_x;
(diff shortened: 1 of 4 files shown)


maxpiepenbrink:
Feel free to shoot this down :) Just ran into this while playing around with a text heavy prototype in MacroQuad where a single draw might be building 30+ of these every frame. The cost is negligible but a & prefix didn't feel much worse than .clone() or `TextParams { ..defaults::Default() }` everywhere.

maxpiepenbrink:
I completely overlooked that many of the other `_ex` functions are also not passing by reference, I have to assume this is just intentional then! 👍 carry on

not-fl3:
I agree, an extra `&` is not a big deal. But, yes, changing all those functions would be too much of a breaking change :(

PR #739, opened 2024-06-16
Fixed render targets not handling depth

yakkudev:
Modified `render_target` so it creates a depth texture for the render pass


profan:
This should probably be optional, otherwise every render target in everyones macroquad projects will now also allocate the depth texture (even if unused)?

yakkudev:
Looks like this has been implemented in b0bf24b.

PR #743, opened 2024-06-21
Added mouse_motion and mouse_wheel to UpdateTrigger

birhburh:
Added mainly for [egui-macroquad to work with blocking_event_loop](https://github.com/birhburh/egui-macroquad/tree/blocking_event_loop)
But it may be convenient for other purposes


not-fl3:
Thanks!

PR #745, opened 2024-06-23
Support public functions in `macroquad::main`

### LINKED ISSUE #626: error: custom attribute panicked
I found a bug, and it might be a bug in macroquad.

My code is like this:

```
#[macroquad::main(window_conf)]
pub async fn gui() {
```

and I'm getting this error:

```
 1  error: custom attribute panicked
    --> night_alerter/src/gui.rs:328:1
     |
 328 | #[macroquad::main(window_conf)]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = help: message: assertion failed: `(left == right)`
               left: `"pub"`,
              right: `"async"` 
```

Sure enough, if I remove `pub`, the error "goes away".

I'm not 100% sure on what exactly `macroquad::main(window_conf)` does. But all the examples in the [documentation](https://docs.rs/macroquad/latest/macroquad/attr.main.html) are private async functions called `main`; I wonder if my entrypoint also needs to be a private async function called main.

My application does also have other things it needs to do, and may or may not end up launching macroquad, which is why I'm not just slapping `#[macroquad::main(window_conf)]` on my `main` function. Maybe I have got the wrong approach.

cyrgani:
Fixes #626 by allowing that the main function with the `macroquad::main` attribute macro is public.

### DIFF
--- a/macroquad_macro/src/lib.rs
+++ b/macroquad_macro/src/lib.rs
@@ -66,6 +66,19 @@ pub fn main(attr: TokenStream, item: TokenStream) -> TokenStream {
             }
         }
     }
+
+    let is_pub = if let TokenTree::Ident(ident) = source.peek().unwrap() {
+        if ident.to_string() == "pub" {
+            // skip 'pub'
+            let _ = source.next().unwrap();
+            true
+        } else {
+            false
+        }
+    } else {
+        false
+    };
+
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

Issue #544, opened 2023-01-01
Request: New Image::set_image_data() to match the existing Image::get_image_data()


Alendri:
## Background of request
Currently `Image::get_image_data()` is defined as `fn(&self) -> &[[u8; 4]]`, which is great. However `Image::update()` is defined as `fn update(&mut self, colors: &[Color])`.

This creates a situation where you are forced to create a `&[Color]` array in order to update the image.

## Suggestion
I would like to request a `Image::set_image_data()` which would be defined as something like `fn set_image_data(&mut self, data: [[u8; 4]])`.
This makes easy to understand "get/set" functions and minimal type changes are forced.

Issue #746, opened 2024-06-23
`Image` fields should be private for soundness


cyrgani:
The following safe code causes undefined behaviour by reading out of bounds:

```rust
fn main() {
    let mut image = macroquad::texture::Image::empty();
    image.width += 1;
    image.height += 1;
    dbg!(image.get_image_data());
}
```

To fix this, all the fields of `Image` should become private, which would be a breaking change. 
The `Image::width` and `Image::height` methods for immutable reading already exist as a replacement, along with `Image::get_image_data`.

Unsafe methods for changing the width and height could be added too if there is a legitimate use case.

PR #748, opened 2024-06-24
Image rework for soundness and usability

### LINKED ISSUE #634: Make Image Safe

cyrgani:
Fixes #634 and #746 by making the fields of image `pub(crate)`, adding the safe constructor from #637 and adding unsafe functions to modify the fields.

This is still work in progress and a breaking change.

TODO: 
* Should `pub(crate)` be removed too? 
* Should the return type be `usize` as before or `u16` as the fields are (or should the fields be `usize` too)? 
* How large is the impact of ths breakage?

### DIFF
--- a/examples/life.rs
+++ b/examples/life.rs
@@ -26,8 +26,8 @@ async fn main() {
     loop {
         clear_background(WHITE);
 
-        let w = image.width();
-        let h = image.height();
+        let w = image.width() as usize;
+        let h = image.height() as usize;
 
         for y in 0..h as i32 {
             for x in 0..w as i32 {
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -98,6 +98,20 @@ impl Image {
         }
     }
 
+    /// Creates an empty Image from a width and height.
+    ///
+    /// ```
+    /// # use macroquad::prelude::*;
+    /// let image = Image::new(32, 32);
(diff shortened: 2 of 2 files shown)


cyrgani:
@not-fl3 I think this is ready for reviewing now.

cyrgani:
Also fixes #544.

cyrgani:
I'm going to split this into some smaller PRs over the next days to hopefully make it easier to review.

PR #751, opened 2024-06-30
add `draw_multiline_text` and `draw_multiline_text_ex`

cyrgani:
This PR adds two functions for drawing multiline text that breaks on newlines. The new `draw_multiline_text` and `draw_multiline_text_ex` functions mirror the existing signatures of the `draw_text` and `draw_text_ex` functions. 

A different option would be to add a field `pub is_multiline: bool` to `TextParams`, which would avoid having two new functions, but that would be a breaking change and would have to wait for 0.5.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -395,16 +395,19 @@ pub fn draw_multiline_text_ex(
     line_distance: Option<f32>,
     params: TextParams,
 ) {
-    let font_line_distance = params
-        .font
-        .map(|font| {
-            font.font
-                .horizontal_line_metrics(1.0)
-                .map(|metrics| metrics.line_gap)
-                .unwrap_or(0.0)
-        })
-        .unwrap_or(0.0);
-    let line_distance = line_distance.unwrap_or(font_line_distance);
+    let line_distance = match line_distance {
+        Some(distance) => distance,
+        None => {
+            let mut font_line_distance = 0.0;
+            if let Some(font) = params.font {
+                if let Some(metrics) = font.font.horizontal_line_metrics(1.0) {
+                    font_line_distance = metrics.line_gap;
+                }
+            }
+            font_line_distance
+        }
+    };
+
     let max_height = text
         .lines()
         .map(|line| measure_text(line, params.font, params.font_size, params.font_scale).height)
--- a/src/text.rs
+++ b/src/text.rs
@@ -362,12 +362,20 @@ pub fn draw_text_ex(text: &str, x: f32, y: f32, params: TextParams) {
     }
 }
 
-/// Draw multiline text with the given font_size and color.
-pub fn draw_multiline_text(text: &str, x: f32, y: f32, font_size: f32, color: Color) {
+/// Draw multiline text with the given font_size, line_distance and color.
+pub fn draw_multiline_text(
+    text: &str,
+    x: f32,
+    y: f32,
+    font_size: f32,
+    line_distance: f32,
+    color: Color,
+) {
     draw_multiline_text_ex(
         text,
         x,
         y,
+        line_distance,
         TextParams {
             font_size: font_size as u16,
             font_scale: 1.0,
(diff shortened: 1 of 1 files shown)


src/text.rs
```diff
@@ -362,6 +362,30 @@ pub fn draw_text_ex(text: &str, x: f32, y: f32, params: TextParams) {
...
+        },
+    )
+}
+
+/// Draw multiline text with custom params such as font, font size and font scale.
+pub fn draw_multiline_text_ex(text: &str, x: f32, mut y: f32, params: TextParams) {
+    for line in text.lines() {
+        let dimensions = measure_text(line, params.font, params.font_size, params.font_scale);
+        draw_text_ex(line, x, y, params.clone());
+        y += dimensions.height + dimensions.offset_y;
```
not-fl3, line 385: should `offset_y` be accumulated for each line? I did not tested it, but I thought `offset_y` is more about where the text is rendered and doesn't affect total line height?

cyrgani: Here are some different approaches I tried now instead of adding the offset:
image

* White is the current version.
* Red is the current version, but without adding `offset_y`.
* Green is this function:
 ```rust
pub fn draw_multiline_text_ex3(text: &str, x: f32, mut y: f32, params: TextParams) {
    let max_height = text
        .lines()
        .map(|line| measure_text(line, params.font, params.font_size, params.font_scale).height)
        .fold(0.0, f32::max);

    for line in text.lines() {
        draw_text_ex(line, x, y, params.clone());
        y += max_height;
    }
}
```
It calculates the biggest height first so that the distance between two adjacent lines is always the same.

* Yellow is the same as Green, but the function takes an extra argument `line_distance: f32` that is added to `y` in each iteration: `y += max_height + line_distance;`.

I think Green or Yellow are preferable over White or Red.

**Comment by not-fl3** [comment]
I do not know enough about fonts to properly comment on this, but, apparently, there are some hints made by font designers for multiline rendering, for example this field looks relevant? https://docs.rs/fontdue/latest/fontdue/struct.LineMetrics.html#structfield.line_gap

**Comment by cyrgani** [comment]
I updated the code to use this fields values if a custom font is given, the font specifies such a distance and the caller doesn't set a line distance manually.

File: src/text.rs [hunk 2]
```diff
@@ -355,13 +355,67 @@ pub fn draw_text_ex(text: &str, x: f32, y: f32, params: TextParams) {
...
+        .font
+        .map(|font| {
+            font.font
+                .horizontal_line_metrics(1.0)
+                .map(|metrics| metrics.line_gap)
+                .unwrap_or(0.0)
+        })
+        .unwrap_or(0.0);
+    let line_distance = line_distance.unwrap_or(font_line_distance);
+    let max_height = text
```
not-fl3, line 408: Doesn't it result in variable lines gap dependning on text content? I don't think that this is how fonts are supposed to work, but, honestly, I did not really researched the topic at all :( 

If you want I can look into this, I think there should be some good references on the topic!

src/text.rs
```diff
@@ -355,13 +355,67 @@ pub fn draw_text_ex(text: &str, x: f32, y: f32, params: TextParams) {
...
+pub fn draw_multiline_text_ex(
+    text: &str,
+    x: f32,
+    mut y: f32,
+    line_distance: Option<f32>,
+    params: TextParams,
+) {
+    let font_line_distance = params
+        .font
+        .map(|font| {
```
not-fl3, line 400: I feel like a few ifs might be a lot more appropriate here, but I would not focus on this until we will figure how to handle multiline gap correctly

src/text.rs
cyrgani: image
This is what the fixed implementation looks like, it should be a constant line gap now.

src/text.rs
cyrgani: Changed it accordingly.

not-fl3: Perfect, thanks for the change!

not-fl3:
Thanks for PR!

PR #757, opened 2024-07-10
fix missing lines bug in `draw_poly_lines`

### LINKED ISSUE #756: draw_poly_lines broken since v0.4.11 (Asteroids example)
I just updated to 0.4.11 and it looks like `draw_poly_lines` is broken somehow.
I just tested it with the asteroid example.

I can confirm it was working in 0.4.8.
I tested it on Windows and MacOS the drawing glitch happens on both platforms.
asteroids

cyrgani:
Removes the private `draw_poly_lines_from_to` function which had misleading argument names and uses `draw_arc` instead as implementation for `draw_poly_lines`.
Fixes #756.

### DIFF
--- a/src/shapes.rs
+++ b/src/shapes.rs
@@ -234,43 +234,6 @@ pub fn draw_poly(x: f32, y: f32, sides: u8, radius: f32, rotation: f32, color: C
     context.gl.geometry(&vertices, &indices);
 }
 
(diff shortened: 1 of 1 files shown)


cyrgani:
CI failures seem unrelated to this and rather related to #754.

not-fl3:
Thanks for PR!

Issue #764, opened 2024-07-20
Arcs don't connect properly


mousetail:
Using `draw_arc` function often doesn't exactly draw to the end of the arc.

Using more segments makes the gap smaller.

It should ideally adjust the segment size such that it can fill the desired angle exactly.

PR #769, opened 2024-07-23
Fix arc exactly matching destination given angle

mousetail:
Before: 
image

After:
image

Short arcs before might not display properly depending on the exact ratio of each step length and the arc span. This adjusts the length of each segment so it will always fit perfectly.

Also uses direct draw calls so the edge is always a straight line to the center of the circle.


mousetail:
Should fix https://github.com/not-fl3/macroquad/issues/764

not-fl3:
Thanks for PR!

PR #770, opened 2024-07-27
fix swapped words in `Color`documentation

### LINKED ISSUE #765: Two docstrings in Color docs are swapped
The docstrings for Color claim that the green channel is blue and vice versa.
https://github.com/not-fl3/macroquad/blob/60e75d20ed393224eaa28b69bd5eb0b02d56eaa9/src/color.rs#L8-L17

cyrgani:
Fixes #765 by swapping the two words.


not-fl3:
Thanks!

PR #772, opened 2024-08-05
added ability to change default texture filter

### LINKED ISSUE #762: Default filter mode for loaded textures
It would be helpful to be able to set the default filter mode for loaded textures/fonts, as setting each texture individually takes a while.

0x177:
resolves #762

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -797,7 +797,7 @@ pub mod conf {
         pub touch: bool,
     }
 
-    #[derive(Default, Debug)]
+    #[derive(Debug)]
     pub struct Conf {
         pub miniquad_conf: miniquad::conf::Conf,
         /// With miniquad_conf.platform.blocking_event_loop = true,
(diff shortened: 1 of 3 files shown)


not-fl3:
LGTM, but maybe add it as a field to `[conf::Conf](https://docs.rs/macroquad/latest/macroquad/conf/struct.Conf.html) as well?

0x177:
[conf::Conf](https://docs.rs/macroquad/latest/macroquad/conf/struct.Conf.html) doesn't seem as accessible as Context, so i do not see how i would access/change the default filter mode if it is stored in Conf

not-fl3:
```
fn window_conf() -> Conf {
    Conf {
        default_texture_filter: Filter,
        ..Default::default()
    }
}

#[macroquad::main(window_conf)]
async fn main() {
    loop {
        clear_background(WHITE);
        next_frame().await
    }
}```

**Comment by 0x177** [comment]
this would require some minor changes to miniquad. i have made a pr for them

**Comment by 0x177** [comment]
my recent commit has implemented it. here is an example of how the api may be used:

```rust
use macroquad::prelude::*;

fn window_conf() -> macroquad::conf::Conf {
    macroquad::conf::Conf {
	miniquad_conf: Conf {
            ..Default::default()
	},
	default_filter_mode: FilterMode::Nearest,
	update_on: None,
    }
}

#[macroquad::main(window_conf)]
async fn main() {
    let texture: Texture2D = load_texture("pixel.png").await.unwrap();
    
    loop {
        clear_background(LIGHTGRAY);
        draw_texture_ex(&texture, 0., 0., WHITE,DrawTextureParams {
	    dest_size: Some(Vec2::new(screen_width(),screen_height())),
	    ..Default::default()
	});
        next_frame().await
    }
}
```

not-fl3:
Thanks for PR!

Issue #775, opened 2024-08-06
Make `draw_character()` and `character_advance()` available to users


bbogdan-ov:
Hi! I'm writing a custom text input, so i need to constantly calculate the size of the text that user enters

Using basic `draw_text()` and `measure_text()` may be quite slow, because it calculates the size for the whole text.
I used slices to calculate the size of the string/text from the beginning to the cursor position, but it also slows down my game!

I want to implement something like this in `text.rs` module:
```rust
// May be called like `CharacterDimensions`
pub struct CharDimensions {
    width: f32,
    offset_y: f32
}

/// Draws the character and returns its size
pub fn draw_character(character: char, x: f32, y: f32, font: Option<&Font>, font_size: u16, color: Color) -> CharDimensions;
/// Returns the character size
pub fn measure_character(character: char, font: Option<&Font>, font_size: u16) -> CharDimensions;
```

Also i would like `draw_text()` and `draw_text_ex()` to return `TextDimensions`, because they have already been calculated, so we don't need to recalculate them again

Want to hear what you think about it :)

PR #776, opened 2024-08-07
Add `TextDimension` return type for `draw_text_ex` and small refactoring

bbogdan-ov:
`draw_text_ex` and `draw_text` now return `TextDimension` to avoid calculating text size using `measure_text` once again

I decided not to add `draw_character` and `measure_character` (as i wanted in my last issue #775), because `draw_text` and `measure_text` do a great job with drawing/measuring characters

Tested with my project (custom text input and other "fancy text")


not-fl3:
Thanks for PR!

PR #779, opened 2024-08-10
graphics: Add Normal vertex attribute

not-fl3:
Also Remove VertexInterop thing for good.

People do ask for extra attributes quite often, and I do not thinkg that slightly larger memory footprint will hurt anyone too much.

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -3,22 +3,38 @@
 use crate::{color::Color, get_context};
 
 use crate::{quad_gl::DrawMode, texture::Texture2D};
-use glam::{vec2, vec3, Quat, Vec2, Vec3};
+use glam::{vec2, vec3, vec4, Quat, Vec2, Vec3, Vec4};
 
+#[repr(C)]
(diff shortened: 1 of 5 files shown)


src/models.rs
```diff
@@ -3,22 +3,38 @@
...
-use glam::{vec2, vec3, Quat, Vec2, Vec3};
+use glam::{vec2, vec3, vec4, Quat, Vec2, Vec3, Vec4};
 
+#[repr(C)]
 #[derive(Clone, Debug, Copy)]
 pub struct Vertex {
     pub position: Vec3,
     pub uv: Vec2,
-    pub color: Color,
+    pub color: [u8; 4],
```
not-fl3, line 13: I wish we could use `color: Vec4` or `color: Color`, it would be one extra general purpose attribute for custom shaders. Unfortunately all current shaders expect to receive `color` in 0..255 range and do this `color / 255.0` thing. Lets wait with this for the next major version bump!

PR #785, opened 2024-08-23
Consistent Camera3D's Fov angle units.

tynberry:
This PR does two changes.

First corrects the Camera3D's documentation for fovy. The docs said it is in degrees, however the code expects radians.

The second one changes the default Camera3D's fovy to be the correct probably intended value. It was 45, which is not normalized radians, thus probably not intended. Now it is 45°. This is a breaking change as the default is different. (By normalizing 45 radians we get something around 58 degrees)


not-fl3:
Thanks for PR!

PR #788, opened 2024-09-03
Add support for tile flip flags

InnocentusLime:
## Synposis

Current `macroquad_tiled` implementation lookups the tileset for a tile by searching via this predicate

```rust
tile >= tileset.firstgid && tile < tileset.firstgid + tileset.tilecount
```

before then converting from `GID` to a tileset-local ID. This is not entirely correct, because the GID is also used to store tile flip flags (https://doc.mapeditor.org/en/stable/reference/global-tile-ids/#tile-flipping).

This causes the above check to fail and all flipped tiles get filtered off.

## What has been done

* When looking up the tileset the tile flip flags are masked off
* Since exposing the raw tile id will be backwards-incompatible they are masked off when constructing the `Tile` struct
* 3 new fields have been added to `Tile` 

## Notes

* I HAVE NOT updated the rendering code, but can if requested
* I have done small local testing and the fix seems to work for me

### DIFF
--- a/tiled/src/lib.rs
+++ b/tiled/src/lib.rs
@@ -30,6 +30,10 @@ pub struct Object {
     pub properties: HashMap<String, String>,
 }
 
+/// Flip operation application order:
+/// 1. flip diagonally
+/// 2. flip horizontally
+/// 3. flip vertically
 #[derive(Debug)]
 pub struct Tile {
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #794, opened 2024-09-05
replace slotmap dependency with reduced local version

cyrgani:
This showed a slight improvement in compile time for me, as `slotmap` is only used in `texture.rs` and most of it is not needed and many things can be simplified based on the limited use case.


not-fl3:
Thanks, great work! Yeah, the whole slotmap deal was a big mistake, but with your fix we at least will not pay for it with compile time.

PR #795, opened 2024-09-06
replace some transmutes with safe pointer casts

cyrgani:
There is no need to use `unsafe` here. 
Note that even after this change, `scene` still contains more unsafe code that the rest of macroquad (51 of 100 occurences of `unsafe` are in scene.rs), so maybe it's time to put `scene` into its own crate and remove it from regular macroquad (will try to prepare a PR later).


not-fl3:
Again, I do not mind mergin this kind of PRs if those changes are somehow useful, but I would rather not touch anything in /exerimental

PR #796, opened 2024-09-06
remove unsound experimental coroutines

### LINKED ISSUE #723: tweens::linear is unsound
The following code segfaults:

```rust
struct Node {
    b: Box<u8>,
    c: Box<u8>,
}
impl scene::Node for Node {}
let node = scene::add_node(Node {
    b: Box::new(0),
    c: Box::new(0),
});
tweens::linear(
    node,
    |node| {
        dbg!(&node.b);
        dbg!(&node.c);
        todo!()
    },
    0.0,
    1.0,
    1.0,
);
```

The `lens` callback receives a mutable reference pointing to uninitialized memory, which is immediate UB.

cyrgani:
The following problematic piece of code in `Handle::lens` is always UB:
```rust
let mut base = std::mem::MaybeUninit::<T>::uninit();
let field = f(std::mem::transmute(base.as_mut_ptr())) as *mut _ as *mut u8;
```
The `MaybeUninit::as_mut_ptr` documentation says:
> Reading from this pointer or turning it into a reference is undefined behavior unless the `MaybeUninit<T>` is initialized.

But `f` takes a mutable reference as argument, so this is UB.

Since I did not see an easy fix for this and it is only an experimental API, it should be fine to just remove it.

Closes #723.


cyrgani:
Closing this since #798 includes this change.

PR #798, opened 2024-09-06
 extract `scene` into separate crate

### LINKED ISSUE #723: tweens::linear is unsound

cyrgani:
As said in #768: 
> Honestly, `scene` overall needs to go. It should be either heavily refactored or be extracted into a separate crate.

This PR does that by removing the `scene` logic from macroquad and only adding a small handler that the new `macroquad-scene` crate can use to register a few functions so that macroquad can interact with the scene as before.
Also moved were the `scene`-related tests, the `state_machine` module and the `CapabilityTrait` derive macro.

A simple implementation of the proposed new `macroquad-scene` crate can be found here: https://github.com/cyrgani/macroquad-scene 

Supersedes #795 and #796 (and likely some more), fixes #723.


not-fl3:
I do not want to touch anything in `scene`/`coroutines` until macroquad-0.5, and macroquad-0.5 is not goint to have any of this anyway.

cyrgani:
Will there be a replacement for it in 0.5?

not-fl3:
I don't think so.
Just as you said, it should be a crate!

cyrgani:
I'll keep this waiting until 0.5 is near, then.

PR #800, opened 2024-09-08
fix several clippy complexity warnings

cyrgani:
A first step towards making clippy complain less: this PR removes unnecessary casts to the same type, unused lifetimes, changes `#cfg(all(feature = "audio"))]` to `#[cfg(feature = "audio")]` and `!CONTEXT.is_none()` to `CONTEXT.is_some()`.

I have not focused on the `style` clippy warnings, as they are often personal preferences, so the changes in this PR should be trivial simplifications only.

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -5,10 +5,10 @@
 use crate::{file::load_file, get_context, Error};
 use std::sync::Arc;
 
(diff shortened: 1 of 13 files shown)


not-fl3:
Thanks for PR!

PR #802, opened 2024-09-10
bump fontdue version

cyrgani:
Should be unproblematic because `fontdue`, unlike `image` and `glam`, is purely an internal dependency and is not publicly reexported.


not-fl3:
Thanks!

PR #806, opened 2024-09-12
fixed gl_set_drawcall_buffer_capacity memory leak

yui-915:
`QuadGl::update_drawcall_capacity()` (which is called by `window::gl_set_drawcall_buffer_capacity`) leaks video memory by not deleting the buffers that were already allocated on the gpu

this could be easily noticed when calling it in a loop (for whatever reason)


not-fl3:
How could I reproduce the leak? I have pretty stable memory on all the examples

not-fl3:
Oh, I see what you mean, sorry for confusion. 

Thanks for PR!

PR #807, opened 2024-09-12
Armotize cpu drawcalls

InnocentusLime:
## Synposis

The cost of breaking the batching in macroquad is too big:
* Each unbatched call creates a `410kb` sized buffer on the CPU
* Each unbatched call creates a new buffer on the GPU

This PR aims to eliminate the big allocations on the CPU side of things to reduce the code of breaking the batching.

The fix is based on the assumption that a ~1mb sized batch is very rare and that most of that 1mb is unused space.

## The implementation

Instead of using a buffer per draw call, the implementation uses a big contigious buffer. Each DrawCall instead stores `(start, len)` in that buffer, essentially acting as a slice

### DIFF
--- a/src/quad_gl.rs
+++ b/src/quad_gl.rs
@@ -20,9 +20,6 @@ pub enum DrawMode {
 pub struct GlPipeline(usize);
 
 struct DrawCall {
-    // vertices: Vec<Vertex>,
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR, great work!

PR #809, opened 2024-09-12
Made drawcalls batching search depth adjustable

yui-915:
drawcalls batching now searches a set depth instead of only looking at the last drawcall 
default depth is 1 so it shouldn't effect anything 
users who want to trade cpu cycles for potentially better batching (therefore more fps) can increase the search depth 
by calling `macroquad::window::gl_set_drawcalls_batching_search_depth()`

this PR aims to solve the altering textures issue described in #805
but more focused on the drawcalls count rather than the allocated memory (it still helps by reducing the buffers count)

this is more of a workaround rather than a proper fix, until a better batching system is implemented

### DIFF
--- a/src/quad_gl.rs
+++ b/src/quad_gl.rs
@@ -558,6 +558,7 @@ pub struct QuadGl {
     draw_calls: Vec<DrawCall>,
     draw_calls_bindings: Vec<Bindings>,
     draw_calls_count: usize,
+    draw_calls_batching_search_depth: usize,
     state: GlState,
     start_time: f64,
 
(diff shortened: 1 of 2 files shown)


InnocentusLime:
I am concerned it might have an effect on the draw order. Let's say we have two draw calls at some point in time:

```
[DC1, DC2]
```

When I do 

```rust
draw_texture(0.0, 0.0, 32.0, 32.0, WHITE, &ferris);
``` 

I expect ferris to be drawn on top of everything else. But if its draw call gets batched to `DC1` -- ferris will be drawn before `DC2`.  

This could be okay for 3D environments as they rely primarily on depth-testing (as far as I am concerned), but may not be perfect for 2D 🤔

yui-915:
apparently, ferris doesn't even get drawn .... 
the recent changes to how drawcalls vertices and indices assumes what the drawcall being modified is the most recent one (prior to this pr, this was always the case) 
but now I broke everything :D 

imma close this as it's probably not worth fixing (will require re-implementing alot of things in quad_gl.rs)

PR #818, opened 2024-09-23
deprecate the `color_u8!` macro and make methods `const`

cyrgani:
With Rust 1.82, `const_fn_floating_point_arithmetic` is stable (see rust-lang/rust#57241).

This means that `Color::from_rgba` can be `const` and the `color_u8!` macro is not required anymore. `Color::from_rgba_f32` was added too since the macro allowed passing `f32` as well as `u8` (or anything else, actually).

Note that this should only be merged after the release of 1.82.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -16,10 +16,12 @@ pub struct Color {
     pub a: f32,
 }
 
-/// Build a color from 4 components of 0..255 values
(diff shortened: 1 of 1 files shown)


cyrgani:
superseded by #918

PR #820, opened 2024-09-23
Fix glyph cache crash

### LINKED ISSUE #815: Bug in the glyph caching system
Hey, running this program

```rust
use macroquad::{
    prelude::*,
    ui::{root_ui, widgets::Button, Skin},
};

#[macroquad::main("a bug")]
async fn main() {
    let font = load_ttf_font("OpenSans-Regular.ttf").await.unwrap();

    let button_style = root_ui()
        .style_builder()
        .with_font(&font)
        .unwrap()
        // .text_color(Color::from_rgba(180, 180, 100, 255))
        .font_size(23)
        .build();

    let button_skin = Skin {
        button_style,
        ..root_ui().default_skin()
    };

    root_ui().push_skin(&button_skin);
    measure_text("I'd love to.", Some(&font), 23, 1.0);
    measure_text("Sorry, I already drunk too much", Some(&font), 23, 1.0);

    loop {
        Button::new("I'd love to.")
            .position(vec2(0., 0.))
            .ui(&mut root_ui());
        next_frame().await;
    }
}
```

will crash with the following error:

```
thread 'main' panicked at /home/felix/.cargo/registry/src/index.crates.io-6f17d22bba15001f/macroquad-0.4.13/src/text.rs:171:53:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I tried to find out what the problem is, and it seems that the entry for the character disappears from the font-atlas for some reason, but not from the characters map. Sadly, I have no idea why. I'll attach the font file.

[OpenSans-Regular.zip](https://github.com/user-attachments/files/17090224/OpenSans-Regular.zip)

It seems the cache is corrupted when called from the buttons ui method, but still ok at the end of the 2nd measure_text call in my code. To me, it looked like the only point in time where something can disappear from the cache, is during a texture update, when the texture needs to be resized, however, as far as I could tell, this didn't happen. Is there any other path to modifying the atlas? I'd be happy to commit a fix for that, but the input from someone more familiar with this code would be useful.

InnocentusLime:
## Synopsis

Fixes #815. The `Style::with_font` method clones the received font and then changes it, placing a different atlas pointer in the cloned font. It, however, doesn't do anything with the `characters` field. 

Because of that, if the original font caches more glyphs -- the font passed to `Style::with_font` ends up in an inconsistent state, which leads to either crashed or garbage looking graphics. This happens from very naive things like measuring text after construing a `Style`, so I think it's worth fixing.

The fix worked when I tested it against the example in issue #815.

## The Fix

I simply modified the routine to supply the cloned font with its own `characters` map. To do this I also had to add a `pub(crate)` method for setting `characters`.

## Alternatives

The alternative is to stop giving the cloned font a different atlas. But it looked like it might be against the original intention behind `Style`'s design.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -77,6 +77,13 @@ impl Font {
         self.atlas = atlas;
     }
 
+    pub(crate) fn set_characters(
+        &mut self,
+        characters: Arc<Mutex<HashMap<(char, u16), CharacterInfo>>>,
+    ) {
+        self.characters = characters;
+    }
+
     pub(crate) fn ascent(&self, font_size: f32) -> f32 {
         self.font.horizontal_line_metrics(font_size).unwrap().ascent
     }
--- a/src/ui/style.rs
+++ b/src/ui/style.rs
@@ -10,7 +10,10 @@ use crate::{
     Error,
 };
 
-use std::sync::{Arc, Mutex};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
 
 pub struct StyleBuilder {
     atlas: Arc<Mutex<Atlas>>,
@@ -60,6 +63,7 @@ impl StyleBuilder {
     pub fn with_font(self, font: &Font) -> Result<StyleBuilder, Error> {
         let mut font = font.clone();
         font.set_atlas(self.atlas.clone());
+        font.set_characters(Arc::new(Mutex::new(HashMap::new())));
         Ok(StyleBuilder {
             font: Arc::new(Mutex::new(font)),
             ..self


InnocentusLime:
I have PR'd the fix. Let me know if it doesn't work :)

KnorrFG:
Hey, I was wondering when this is getting merged. Are there any policies in place that prevent the merge?

InnocentusLime:
@not-fl3 

Is there anything preventing the merge of this PR? 🤔

not-fl3:
Sorry for delay! 

Thanks for PR!

KnorrFG:
Thank you for fixing this :)

PR #822, opened 2024-09-24
use shorthand initialization and remove redundant variables

cyrgani:
Followup to #817: remove redundant local variables, wrapper closures, explicit returns, unneded early variable declarations and unneded (mutable) references.


not-fl3:
Thanks!

PR #824, opened 2024-09-24
Overwrite render pass to Some/None instead of just setting it to `Some`

oli-obk:
https://github.com/not-fl3/macroquad/pull/719 changed this to only ever set the render target to Some if the camera's is Some, but never unset it to None.

In my game I currently need to reset with `set_camera_default()` before setting a camera that doesn't have a render target set.

### DIFF
--- a/src/camera.rs
+++ b/src/camera.rs
@@ -244,9 +244,11 @@ pub fn set_camera(camera: &dyn Camera) {
     // flush previous camera draw calls
     context.perform_render_passes();
 
-    if let Some(render_pass) = camera.render_pass() {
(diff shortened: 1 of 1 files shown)


profan:
Seems like this was fixed in [4ad23c8](https://github.com/not-fl3/macroquad/commit/4ad23c8bbb4e9a2a0632fafce1fb60d36af6598c), can probably be closed now :)

(also sorry :eye:)

oli-obk:
Ah great, thanks!

not-fl3:
Ohh shame on me, solution was there in PRs all along

PR #829, opened 2024-09-27
macroquad-tiled: Support tile flipping & rotation in `Map::draw_tiles`

Adjective-Object:
This change adds support for `tiled`'s flipping flags when rendering a tilemap

- Adds new api in Map `.spr_flip`, which renders a tile including the tile's [flipping parameters](https://doc.mapeditor.org/en/stable/reference/global-tile-ids/#mapping-a-gid-to-a-local-tile-id)
- Uses that api in .draw_tiles

From a simple test map I made:

| Before | After |
| ------ | ----- 
| image|image|

### DIFF
--- a/tiled/src/lib.rs
+++ b/tiled/src/lib.rs
@@ -7,6 +7,7 @@ use std::collections::HashMap;
 mod error;
 mod tiled;
 
+use core::f32::consts::PI;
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks!

PR #832, opened 2024-09-29
Update documentation reference for TextDimensions to fix #599.

### LINKED ISSUE #599: TextDimensions docs mention not existant "text_dimentions" example
https://docs.rs/macroquad/latest/macroquad/text/struct.TextDimensions.html
has this line:
“For reference check “text_dimensions” example.”

I think you meant the “Text measures”

profan:
Fixes a reference in docs to a non-existent example.


not-fl3:
Thanks!

PR #838, opened 2024-10-11
Fix default filter mode ignored by from_rgba8

### LINKED ISSUE #830: Texture2d::from_rgba8 ignores default_filter_mode
## The problem

`Texture2d::from_rgba8` and, consequently, `Texture2d::from_image` completely ignore the `default_filter_mode` setting.

## Reproducible Example

The following code shows the example

```rust
#[macroquad::main("Blurry")]
async fn main() {
    set_default_filter_mode(FilterMode::Nearest);

    let t = Texture2D::from_rgba8(3, 3, &[
        255, 255, 255, 255,    0  , 0  , 0  , 0  ,  255, 255, 255, 255,
        0  , 0  , 0  , 0  ,    255, 255, 255, 255,  0  , 0  , 0  , 0  ,
        255, 255, 255, 255,    0  , 0  , 0  , 0  ,  255, 255, 255, 255,
    ]);
    /* Uncomment to get the intended result */
    // t.set_filter(FilterMode::Nearest);

    loop {
        clear_background(GRAY);

        draw_texture_ex(
            &t,
            0.0,
            0.0,
            WHITE,
            DrawTextureParams {
                dest_size: Some(vec2(400.0, 400.0)),
                ..Default::default()
            }
        );

        next_frame().await;
    }
}
```

## The Fix

I can work on the fix if it that behaviour is unintended. I believe it's pretty simple:

* All Texture2d creation stuff eventually boils down to calling `Texture2d::from_rgba8`
* To leverage the delegation pattern, one should simply move `.set_filter` call from `from_file_with_format` to `Texure2d::from_rgba8`

juliohq:
Fixed #830.

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -666,8 +666,6 @@ impl Texture2D {
 
         let ctx = get_context();
 
-        t.set_filter(ctx.default_filter_mode);
-
         t
     }
 
@@ -703,6 +701,7 @@ impl Texture2D {
         let ctx = get_context();
         let texture = ctx.textures.store_texture(texture);
         let texture = Texture2D { texture };
+        texture.set_filter(ctx.default_filter_mode);
 
         ctx.texture_batcher.add_unbatched(&texture);
 


not-fl3:
Thanks for PR!

PR #841, opened 2024-10-14
Tweak some comments and add a new `from_hex_alpha` function

addiment:


.gitignore
```diff
@@ -1,3 +1,38 @@
-/target
-**/*.rs.bk
+# Generated by Cargo
```
not-fl3, line 1: This sneaked into the PR by accident, right? :) Could you please rebase it out?

src/color.rs
```diff
@@ -131,6 +131,25 @@ impl Color {
...
+    /// ```
+    /// use macroquad::prelude::*;
+    ///
+    /// let light_blue = Color::from_hex_alpha(0x3CA7D5, 0x80);
+    /// assert_eq!(light_blue.r, 0.23529412);
+    /// assert_eq!(light_blue.g, 0.654902);
+    /// assert_eq!(light_blue.b, 0.8352941);
+    /// assert_eq!(light_blue.a, 0.5019608);
+    /// ```
+    pub fn from_hex_alpha(hex: u32, alpha: u8) -> Color {
```
not-fl3, line 147: I do not have very strong opinion here, but it looks like `#rrggbbaa` notation is fairly common. Maybe we should follow it to? Keep `from_hex`  `#rrggbb` for the sake of backwards compatibility, while the new one will be `0xRRGGBBAA`?

addiment: oops-- totally forgot that 32 bits has room for alpha. Will patch that in, stupid oversight on my part.

not-fl3: It got room, the problem is that `0xff0000` is either red in `from_hex` or very transparent green in `from_hex_alpha`, which might be confusing. 0xAARRGGBB is a bit less error prone, but with #rrggbbaa being popular - even more confusing. So `hex: u32, alpha: u8` actually does makes sense to me!

PR #842, opened 2024-10-14
respect viewport in Camera::screen_to_world

rileylyman:
I noticed that if you have a custom viewport set on your `Camera2D`, then `screen_to_world` does not work correctly because it always assumes the viewport is `(0, 0, screen_width(), screen_height())`. This change checks if there is a viewport and, if so, uses that instead. Let me know if you want this change; I've just had it in my local copy for a bit now.


not-fl3:
Good catch, thanks for PR!

PR #850, opened 2024-10-20
make all methods `const` that can be `const` on 1.70.0

cyrgani:
This is better than #818 because it compiles sucessfully on rust 1.70.0, which is the MSRV due to miniquad using `OnceLock`, which is stable since then.


not-fl3:
Thanks for PR!

PR #852, opened 2024-10-26
Add `draw_fps` helper function

juliohq:
This PR adds `draw_fps` and `draw_frame_time` functions for drawing the result of `get_fps` and `get_frame_time`, respectively.

The rustaceanmark example was updated to use the newly added `draw_fps` function.

### DIFF
--- a/examples/rustaceanmark.rs
+++ b/examples/rustaceanmark.rs
@@ -55,7 +55,7 @@ async fn main() {
             );
         }
 
-        draw_text(format!("FPS: {}", get_fps()).as_str(), 0., 16., 32., WHITE);
+        draw_fps();
         draw_text(
             format!("Rustaceanes: {}", rustaceanes.len()).as_str(),
             0.,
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks for PR! 
I was thinking about `draw_fps`, and I think that in a shape of `draw_fps(0., 16., 32., WHITE);` it doesn't really provide enough value to justify an existance of a separate function. And I don't really think anyone would ever use draw_frame_time, and if they will need it - I would say its too niche of the use case for a special function.

However, `draw_fps()` with no arguments might be usefull for quick tests, what do you think about this?

juliohq:
Yes, I think it makes sense to remove `draw_frame_time` in this case and add the doc-comment you mentioned to `draw_fps`.

not-fl3:
I would rather have `draw_fps()` as a no-argument function. If custom format, position, font size of anything is required - its not that much of a problem to call `draw_text` I belive.

juliohq:
>I would rather have draw_fps() as a no-argument function.

Sure, it should work better like that I think.

not-fl3:
Perfect, thanks for PR!

PR #860, opened 2024-11-17
Reset viewport in `set_default_camera`

agerasev:
Hi, thank you for such an awesome crate!

Currently `set_default_camera` does not reset viewport, so the following workaround needs to be used:

```rust
// Workaround to reset viewport
set_camera(&Camera2D::default());

set_default_camera();
```

This pull request fixes that.


not-fl3:
Thanks for PR!

PR #867, opened 2024-12-18
warn when calling broken functions that invoke UB

cyrgani:
As explained in #723, these functions are unsound, so lets warn users that use them.


oli-obk:
It's in the experimental module. I think it can just be fixed by changing the API in ways that make it safe

PR #882, opened 2025-01-12
add help message for `#[macroquad::main]` on a non-async function

cyrgani:
Just print an extra message if the `async` token is not found, similar to the case below in the code.


not-fl3:
Thanks for PR!

PR #883, opened 2025-01-12
Add wrap_text and an example

JaniM:
Adds a new `wrap_text` function that calculates line breaks such that the given string would fit in the given line length. The example should be quite self-evident.

PR #884, opened 2025-01-13
add measure_multiline_text

JaniM:
Expands on top of https://github.com/not-fl3/macroquad/pull/883, that should be merged first. This PR only adds the latest commit so review should focus on that.

This change adds a new function, `measure_multiline_text` and updates `draw_multiline_text` to return the dimensions. I expanded the `text_wrap` example to show its results.


amatgil:
Are there any issues blocking this PR? (I'd be happy to fix them)

I've come across the exact same situation: `draw_text` has its `measure_text` counterpart, but `draw_multiline_text` does not have `measure_multiline_text` (and it doesn't return the dimensions, like `measure_text` does).

(It could be implementable by the user, I think, but the `font` attribute in `Font` is private so, the font's vertical spacing is inaccessible)

Issue #663, opened 2023-10-22
Allow files to be drag and dropped into game window


9D-Tony:
It would be a useful feature to have the drag and drop in macroquad that when files are dropped into the window then it gives you a vector of filenames back that have been dropped into the window.

This action is supported in other libraries like [raylib](https://docs.rs/raylib/latest/raylib/ffi/fn.GetDroppedFiles.html) but only for supported platforms e.g Windows, Linux, OSX & Web

Nrosa01:
I support this, it's something really useful for testing or adding an easy way to add mods and stuff

zyllian:
Note that this requires support from miniquad for each platform, notably I believe this is still not implemented on macos.

PR #889, opened 2025-01-24
Implement drag and drop file support

zyllian:
Adds drag and drop file support to Macroquad for #663. Note that this will only work through WASM until Miniquad supports drag and drop on other platforms. (I've just submitted a PR to add support for this on Windows to MIniquad: https://github.com/not-fl3/miniquad/pull/518)

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -2,10 +2,10 @@
 
 use std::collections::HashSet;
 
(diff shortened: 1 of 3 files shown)


not-fl3:
Thanks for PR!

PR #891, opened 2025-01-29
Color fixes/utility method

jonatino:
Small enhancement for Color, and new method to easily mutate colors in constants


not-fl3:
Thanks for PR!

PR #900, opened 2025-02-13
Make the texture batcher respect default mode

InnocentusLime:
## Synopsis

The current implementation of `build_texture_atlas()` ends up making an atlas, that ignores the default filtering mode. This PR remedies this situation by making `set_default_filter_mode()` also set the filter mode of the texture atlas.

## Alternatives

The alternative to this fix is to introduce a new piece of API -- `set_texture_atlas_filtering()`


not-fl3:
Good catch, thanks for PR!

PR #901, opened 2025-02-13
Added draw_billboard to models.rs

Barysk:
I think having a simple billboard model might be a useful feature.


mangusek (commented):
Nice idea. Billboards are very useful in 3D games

PR #903, opened 2025-02-16
Updates the position of the checkbox relative to the window

### LINKED ISSUE #902: [UI] checkbox position not relative to the window its passed to
the checkbox position is not relative to the sub-window (not the game window) when placed inside a root_ui.().window() window
i spent hours trying to see why the checkbox disappears when i add a custom position relative to the sub-window.

finally debugged the library checkbox.rs and found that the default position when i did not pass a position (when pos is None) is vec2(252.0, 197.0) <= from context.window.cursor.fit(size, Layout::Vertical) in my project. so logically, the position set here should be wrt to context.window.cursor.fit(size, Layout::Vertical) but it's not right now, I will try to open a PR with a fix for this.

Ashfaaq18:
fixes #902 

with this fix, now the checkbox shows up and responds with the position relative to window its in.

### DIFF
--- a/src/ui/widgets/checkbox.rs
+++ b/src/ui/widgets/checkbox.rs
@@ -64,6 +64,7 @@ impl<'a> Checkbox<'a> {
 
         let pos = self
             .pos
+            .map(|pos| pos + context.window.cursor.fit(size, Layout::Vertical))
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #907, opened 2025-02-19
Updates position of InputText to be relative to parent window is spec…

### LINKED ISSUE #906: [UI] InputText and EditBox don't position themselves relative to the window its passed to
Was trying to position a few things, it wasn't working, looked at the source code and noticed the issue. 
I checked to see if there was a pull request or issue about this already out there, #902 seems to be about the same thing. 
I will submit a PR implementing the fix in PR #903 to both InputText and EditBox

camilomcatasus:
Partially fixes #906 am open to continuing work on PR to make Editbox position itself relatively as well despite the issues I mentioned.

### DIFF
--- a/src/ui/widgets/input.rs
+++ b/src/ui/widgets/input.rs
@@ -92,6 +92,7 @@ impl<'a> InputText<'a> {
 
         let pos = self
             .pos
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #911, opened 2025-02-26
Fix backtrace dependency

madsmtm:
`backtrace` had a breaking change in https://github.com/rust-lang/backtrace-rs/pull/615, where it removed old feature flags. This forces Cargo to fall back to an older version of `backtrace` (namely `v0.3.71`).


not-fl3:
Thanks for PR!

Issue #621, opened 2023-07-19
Loading fonts causes panic


hasikill:
I encountered a panic while loading "Microsoft YaHei.ttf"
```
thread 'main' panicked at 'Vertical fonts are not supported', C:\Users\pc\.cargo\registry\src\index.crates.io-6f17d22bba15001f\macroquad-0.3.26\src\text.rs:92:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

whouishere:
just tested this and the mentioned font seems to load and work with no problems when the panic on [text.rs:83](https://github.com/not-fl3/macroquad/blob/a84eb77572c20e4757a8ece517cdcd1e5bd20593/src/text.rs#L83), is commented out, even with chinese characters.

@not-fl3 any reasons this panic should be kept? any other vertical fonts that don't work?

not-fl3:
> just tested this and the mentioned font seems to load and work with no problems when the panic on [text.rs:83](https://github.com/not-fl3/macroquad/blob/a84eb77572c20e4757a8ece517cdcd1e5bd20593/src/text.rs#L83), is commented out, even with chinese characters.

I do not not enough about vertical fonts, but  I assume that they should be rendered vertically (`draw_text` should draw a label with each symbol above/beyound each other, not left to right).

So I would keep this panic on until the correct vertical rendering would be fixed.

TrolledWoods:
I don't know enough about fonts to say too much about this, but my guess is that `advance_height` being non-zero doesn't mean the font is "vertical", it just means that it has the potential to be rendered vertically. Chinese characters are quite commonly used in both horizontal and vertical writing, so it makes sense that a font would try to support both styles. So I think ignoring it, or maybe logging a warning or something, is preferable to panicking here (a lot of chinese/japanese fonts have advance_height set, so it's pretty much impossible to render even just horizontal text in those languages using macroquad right now without forking)

neunenak:
I'm running into this issue trying to load an ordinary Noto Sans CJK font from my system's font path - trying to open the .ttc file at all causes this panic, even though I want to render fonts horizontally.

djtuBIG-MaliceX:
Bit annoying to have to vendor and invasively modify the source code just to be able to use NotoSansJP on here.  There's no reason to panic this if the API only supports left-to-right writing to begin with.

djtuBIG-MaliceX:
I have raised #913 to address at least my current use case for vertical font rotation.

PR #913, opened 2025-03-09
Fix multi-line text rendering for rotated text after first line

djtuBIG-MaliceX:
# WHAT IS THIS
This addresses an issue where `draw_multiline_text_ex()` would incorrectly keep subsequent lines 'vertically flush' with the first line, when it should also rotate horizontally in order to given the visual presentation of a fully multi-line rotated text block on a **single render**.  Below screenshot is an example of what it should look like with a single draw, and yes I've tested this with animated spinning text.

Correct output:
image

This also removes the restriction of not supporting vertical fonts, as by default it should use the font size given as the offset.  (Tested with [NotoSansJP](https://fonts.google.com/noto/specimen/Noto+Sans+JP)).  Ideally, usage of `advance_height` would be handled as a separate `draw_text_vertical()` function or otherwise some kind of boolean flag. (not part of this PR).

Related:  #621

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -102,9 +102,9 @@ impl Font {
 
         let (metrics, bitmap) = self.font.rasterize(character, size as f32);
 
-        if metrics.advance_height != 0.0 {
-            panic!("Vertical fonts are not supported");
-        }
+        // if metrics.advance_height != 0.0 {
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #916, opened 2025-03-19
make `Color::from_hex` const

cyrgani:
This is very useful because it allows us to write 
```rust
const WHITE_GREEN: Color = Color::from_hex(0xC2FFB9);
```


not-fl3:
Thanks!

PR #919, opened 2025-03-23
make all `Rect` methods `const`, fix a small bug in `Rect::contains`

cyrgani:
Thanks to https://github.com/rust-lang/rust/pull/129195, all methods on `Rect` can become const now, including those with `&mut self`.
Also fixes a small bug where points on the border of a rect where not recognized to be in it, and adds a simple test for that.


not-fl3:
Thanks!

PR #923, opened 2025-03-24
add `set_default_font` and a small example

cyrgani:
As a complement to `get_default_font`, this PR adds `set_default_font` as well as a small example to demonstrate that it works:

image

This makes it much easier to set the default font once before the main loop and then use the regular text drawing APIs, rather than having to explicitly use `draw_text_ex` with the font each time.


not-fl3:
Thanks for all the PRs! :+1:

PR #925, opened 2025-03-24
simplify the reduced `slotmap` even further

cyrgani:
I think that there is no need to store the `version` of the texture, which then makes several types simple wrapper types that can be removed too.
While I tested this for multiple examples and saw no changes in behaviour, I'm not fully sure yet that this does not change the behaviour in every case.


cyrgani:
closing as the potential incorrectness is not worth the small code reduction

PR #926, opened 2025-03-26
replace all `Into` implementations with `From` implementations

cyrgani:
This allows more ergonomic conversions between colors, `f32` arrays and `u8` arrays, such as:
```rust
Color::from([16, 25, 116, 34]);
```
The `Into` implementations will be provided automatically, so all existing code continues to work.

The diff can also be reproduced with `cargo clippy --fix --lib -p macroquad`.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -45,31 +45,31 @@ fn color_from_bytes() {
     );
 }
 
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks!

PR #928, opened 2025-03-26
add `Image::as_app_icon` to allow ergnomically setting the window icon

cyrgani:
This PR adds the function `Image::as_app_icon`, which can be used to construct a `miniquad::conf::Icon` for use in a `Conf`. Previously, there was no practical way to do this, except to manually use `include_bytes!` and have three differently scaled versions. (best example of this: https://github.com/not-fl3/miniquad/blob/master/src/default_icon.rs)
The PR also includes a small example that shows it is working as intended.

Note that this is built on top of #927.

### DIFF
--- a/examples/custom_app_icon.rs
+++ b/examples/custom_app_icon.rs
@@ -1,0 +1,21 @@
+use macroquad::prelude::*;
+use std::fs::read;
+
+fn conf() -> Conf {
+    let image =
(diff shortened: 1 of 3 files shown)


cyrgani:
An alternative would be to merge not-fl3/miniquad#540 first and then either close this PR or make `Image::as_app_icon` a trivial wrapper around `Icon::from_bytes`.

PR #929, opened 2025-03-27
stop flipping the image in `Image::export_png`

### LINKED ISSUE #987: Images are inverted on the Y axis
I've created a small example bit of code with:

``` rust
use std::fs;
use macroquad::prelude::*;

fn main() {
    let bytes = fs::read("input.png").unwrap();
    let image = Image::from_file_with_format(&bytes, Some(ImageFormat::Png)).unwrap();
    image.export_png("output.png");
}
```

Running this will result in the output image being flipped vertically - this seems incorrect to me.

---

`input.png`

Image

`output.png`

Image

---

System info:

```
$ cargo --version
cargo 1.83.0 (5ffbef321 2024-10-29)
$ rustc --version
rustc 1.83.0 (90b35a623 2024-11-26)
$ uname
Darwin
$ cat Cargo.toml
[package]
name = "macroquad-flip-image"
version = "0.1.0"
edition = "2021"

[dependencies]
macroquad = "0.4.14"
```

cyrgani:
I don't know why this was added in the first place. When I tried using this method, I had to remove the flipping code to actually generate the right image.
Maybe it was a remnant of a previous refactoring that was not removed when it was no longer needed?
Fixes #987.

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -304,19 +304,9 @@ impl Image {
     /// Saves this image as a PNG file.
     /// This method is not supported on web and will panic.
     pub fn export_png(&self, path: &str) {
-        let mut bytes = vec![0; self.width as usize * self.height as usize * 4];
-
-        // flip the image before saving
-        for y in 0..self.height as usize {
-            for x in 0..self.width as usize * 4 {
-                bytes[y * self.width as usize * 4 + x] =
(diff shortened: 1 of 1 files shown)


pkptzx:
This cannot be merged. At the very least, you should add a new parameter or a separate method. The reason is that saving the entire window to an image relies on this very same function: get_screen_data().export_png(), which is correct. However, in other cases, Macroquad's Image is almost always created for textures

PR #938, opened 2025-04-05
apply `clippy::question_mark` lint

cyrgani:
Just another small code cleanup to replace 
```rust
if foo.is_none() {
    return None;
}
foo.unwrap()
```
with 
```rust
foo?
```
in several places.


not-fl3:
I would rather not touch anything under /experimental, but if helps somehow - I can merge it, np!

PR #939, opened 2025-04-05
make circles look more like circles

cyrgani:
Circles are currently 20-side polygons. This PR increases that number to 100 for much better visual effect. It also fixes an inconsistency: previously, circle outlines were 30-side polygons, but filled circles had just 20 sides.

### Before
image

### After
image

Testing this on an example project revealed that the old version took around 3µs per circle in release mode, while the new one took 4µs. Given that the circle is five times more precise now, that seems like a good tradeoff.

This supersedes #713.

### DIFF
--- a/src/shapes.rs
+++ b/src/shapes.rs
@@ -250,12 +250,12 @@ pub fn draw_poly_lines(
 
 /// Draws a solid circle centered at `[x, y]` with a given radius `r` and `color`.
 pub fn draw_circle(x: f32, y: f32, r: f32, color: Color) {
-    draw_poly(x, y, 20, r, 0., color);
+    draw_poly(x, y, 100, r, 0., color);
 }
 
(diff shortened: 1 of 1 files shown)


not-fl3:
It always will be a compromise, people do use draw_circle to draw a thousand of very little circles, and 5 would be a good enough number of vertices. And people will use draw_circles for full_screen circles, where even a 100 will be not enough. 

20 looks like a decent middle ground tbh

cyrgani:
For every use case, people can just call `draw_poly` and set whatever amount of sides they want. So I think it is just the question which use case for circles is more common and should be covered by `draw_circle`: many very small ones or fewer bigger ones. I don't know which one is more common though, so I see why you don't want to change this.

Maybe there is some better solution that calls into miniquad rendering directly without needing `draw_poly` and without speed losses?

PR #944, opened 2025-04-10
Set and Get Actor Size

danwritecode:
Hey there! I'm working on a game right now using Macroquad and it's been great! However, I ran into the need to implement dynamically sized hit boxes based on various animation states (crouching for example). 

After reading the source code for the platformer lib, I realized that there needed to be some methods to set and get the actor size. 

Regarding "setting": to do this correctly, the actors position also needs to be changed based on the new height passed in otherwise the hit box would do weird things. You'll see that in the implementation. 

Let me know what you think!


not-fl3:
Thanks for PR!

PR #956, opened 2025-04-23
Make the ToBytes trait unsafe

meithecatte:
Because of the `ToBytes` impls on arrays and slices, which do not make use of `to_bytes` of the underlying type, implementing `ToBytes` makes a promise to unsafe code. This should be explicit, by having the trait itself be unsafe.

Also replaces the impl for `&[T]` with one for `[T]`, to avoid depending on the layout of `&[T]`. This is, technically, a breaking change – however, the code that would break didn't actually work at runtime in the first place.

### DIFF
--- a/src/tobytes.rs
+++ b/src/tobytes.rs
@@ -1,11 +1,16 @@
 use crate::math::{Mat4, Vec2, Vec3, Vec4};
 
-pub trait ToBytes {
+/// # Safety
+///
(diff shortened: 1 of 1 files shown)


KurlykovDanila:
In your opinion, how public should this trait actually be? I looked through the library, and this trait is only used in the ``set_uniform`` functions. Maybe it should just be made unavailable for external implementation?

PR #963, opened 2025-05-07
replay window minimized and restored events

### LINKED ISSUE #685: EventHandler not receiving events
```
use crate::miniquad::EventHandler;
use macroquad::prelude::utils::repeat_all_miniquad_input;
use macroquad::prelude::utils::register_input_subscriber;

impl EventHandler for MyEventHandler {
    fn update(&mut self) {
        // Update logic here
    }

    fn draw(&mut self) {
        // Drawing logic here
    }

    fn window_minimized_event(&mut self) {
        panic!("Never hit");
    }

    fn window_restored_event(&mut self) {
        panic!("Never hit");
    }
}

let subscriber = register_input_subscriber();
let mut event_handler = MyEventHandler;
```

game loop:
`repeat_all_miniquad_input(&mut event_handler, subscriber);`

The event handler does not seem to receive events even when the window is opened/closed or minimized. On windows 11. Reproduced on blank first person preview example.

InnocentusLime:
Closes #685

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -273,6 +273,8 @@ enum MiniquadInputEvent {
         x: f32,
         y: f32,
     },
+    WindowMinimized,
+    WindowRestored,
 }
 
 impl MiniquadInputEvent {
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #972, opened 2025-05-17
texture: add extra asserts to Image

### LINKED ISSUE #746: `Image` fields should be private for soundness

InnocentusLime:
Part of the effort to patch the Rust-sec advisory report (https://rustsec.org/advisories/RUSTSEC-2025-0035.html)
Closes #746 by introducing an assert into the `unsafe` functions to avoid UB.

* `Image` already uses the `assert!()` liberally to check its invariants
* This will not harm the performance too much, as the user can always cache the reference

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -169,6 +169,7 @@ impl Image {
     /// Returns this image's data as a slice of 4-byte arrays.
     pub fn get_image_data(&self) -> &[[u8; 4]] {
         use std::slice;
+        assert!(self.width as usize * self.height as usize * 4 == self.bytes.len());
 
         unsafe {
             slice::from_raw_parts(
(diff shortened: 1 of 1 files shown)


InnocentusLime:
Closes #634 
(I think)

InnocentusLime:
@not-fl3 
For some reason #634 has not closed automatically

PR #975, opened 2025-05-20
Added draw_text_pro()

Thepuppetqueen57:
Just draw_text() but with a custom font. It's more useful than it looks btw.

name could maybe be improved idk


profan:
Surely [draw_text_ex](https://docs.rs/macroquad/latest/macroquad/text/fn.draw_text_ex.html) already does exactly this?

Issue #924, opened 2025-03-24
Function 'render_target' makes Wasm application crash since version 0.4.14.


jsgonsette:
The function 'render_target' makes my Wasm application crash when called.
I get this error in the Browser console:
```
app.wasm:0xf089d Uncaught RuntimeError: unreachable in ...
```

This problem does not appear in previous version (0.4.13).

CristianPi:
> The function 'render_target' makes my Wasm application crash when called. I get this error in the Browser console:
> 
> ```
> app.wasm:0xf089d Uncaught RuntimeError: unreachable in ...
> ```
> 
> This problem does not appear in previous version (0.4.13).

i had the same problem, it seems that the js bundle is not kept in sync and we need webgl2 (at least i need it for the readBuffer)

try this in the meantime:

[mq_js_bundle.js](https://jsfiddle.net/gsbqj9rt/)

CristianPi:
> Can you share the setup for reproduction? Where did you acquire the `mq_bundle`? Does the `post_processing` example not work for you too?

the bundle

```bash

#!/bin/bash

function wrap_js {
    echo "(function () {" >> mq_js_bundle.js
    cat $1 >> mq_js_bundle.js
    echo "}());" >> mq_js_bundle.js
}
cat ../../miniquad/js/gl.js > mq_js_bundle.js
wrap_js ../../quad-snd/js/audio.js
wrap_js ../../sapp-jsutils/js/sapp_jsutils.js
wrap_js ../../quad-net/js/quad-net.js
```

you will need to git clone all the other projects (miniquad, quad-snd, sapp-jsutils, quad-net).

**how to test**

Run build-wasm.sh

Edit dist/index.html -> **assets/mq_js_bundle.js** or **https://not-fl3.github.io/miniquad-samples/mq_js_bundle.js**

Launch the index.html with the updated bundle or the web one

[Project](https://github.com/user-attachments/files/19510893/Archive.tar.gz)

I also noticed that gl.readBuffer is used on a webgl1 context (it does not exist, and it crash).
I don't know if that's related to render_target, but i need to use webgl2 to make the code work.

```rust

fn window_conf() -> Conf {
    Conf {
        window_title: "Shader Optimized Universe".to_owned(),
        window_width: 1920,
        window_height: 1080,
        fullscreen: true,
        platform: Platform {
            webgl_version: WebGLVersion::WebGL2,
            ..Default::default()
        },
        ..Default::default()
    }
}
```

[...]

PR #984, opened 2025-07-03
Updated mq_js_bundle.js using attached instruction.

indieleet:
Updated version of mq_js_bundle.js, now I have my wasm builds working correctly. 
Issue #924 may be fixed, but further testing is required.


not-fl3:
Thanks!

Issue #995, opened 2025-08-26
Rect.contains() only works properly with positive width and height


CrusaderTek:
if a rectangle has an origin other than the top left it breaks the contains function. I wrote up a quick and dirty work around:

`pub fn true_contains(rect: &Rect, point: Vec2) -> bool{
    let top: f32;
    let bottom: f32;
    let left: f32;
    let right: f32;
    if rect.h < 0.0{
          top = rect.y + rect.h;
          bottom = rect.y;
    }
    else{
         top = rect.y;
         bottom = rect.y + rect.h;
    }
    if rect.w < 0.0{
         left = rect.x + rect.w;
         right = rect.x;
    }
    else{
         left = rect.x;
         right = rect.x + rect.w;
    }
    if point.x >= left
    && point.x < right
    && point.y < bottom
    && point.y >= top{
        true
    }
    else{
        false
    }
}`

oli-obk:
This can be closed now that you can normalize Rects easily I think?

PR #1000, opened 2025-10-22
add normalized function for Rect struct - 995

michaelcortese:
#995 

in cases where there is a Rect with a negative width or height, `Rect::normalized(&self)` returns a Rect with a positive width and height, adjusting the corner back to the top left


adrylain:
Please merge this.

Issue #1002, opened 2025-10-30
RenderTarget Lifetime Issue. Should Be Deferred for Deletion Like Textures


jonatino:
Spent a few days tracking this one down. Really hard to find in a big codebase. Only figured it out thanks to this Discord message:
https://discord.com/channels/710177966440579103/710180051349405746/1345439660222054400

Currently, when using render targets, it’s up to the developer to make sure the target outlives the current frame, which is pretty easy to mess up, especially in larger projects.

Would it make sense to handle render targets like textures and schedule them for deletion instead of dropping them right away?

From what I understand, textures are garbage collected at the end of the frame, so it’d be nice if render targets had that same behavior.

Where as RenderPass (which owns the textures/render target) is dropped right away including all of its textures:
```rust
impl Drop for RenderPass {
    fn drop(&mut self) {
        if Arc::strong_count(&self.render_pass) < 2 {
            let context = get_quad_context();
            context.delete_render_pass(*self.render_pass);
        }
    }
}
```

Reproduction below for the bug (press space to toggle between dropping the render target early vs keeping it alive).

The expected output is for there to always be an big blue rectangle with a smaller orange one inside it, with a white X over it. Currently, this is only the case if we manually hold on to render target reference until the frame is over.

```rust
use macroquad::prelude::*;

const RT_W: u32 = 128;
const RT_H: u32 = 128;

#[macroquad::main("RenderTarget lifetime repro")]
async fn main() {
    let mut broken_mode = true;

    loop {
        clear_background(DARKGRAY);

        if is_key_pressed(KeyCode::Space) {
            broken_mode = !broken_mode;
        }

        // BROKEN: RenderTarget is created and dropped inside helper (before frame cleanup)
        let _rt = if broken_mode {
            // Intentionally drop the RenderTarget before next_frame()
            draw_render_target();
            None
        } else {
            let rt = draw_render_target();
            Some(rt)
        };

        let color = if broken_mode { RED } else { GREEN };
        let text = if broken_mode { "BROKEN: render target dropped before next_frame()" } else { "WORKING: render target kept alive until next_frame()" };
        draw_text(format!("{text}\nbackground should be black, with a blue square with a orange square\ninside it and a white X over it (press SPACE to toggle modes)").as_str(), 16.0, 28.0, 22.0, color);

        next_frame().await;
        // _rt drops here, after frame cleanup
    }
}

fn draw_render_target() -> RenderTarget {
    let rt = render_target(RT_W, RT_H);

[...]

PR #1003, opened 2025-10-30
fixes(#1002): Schedule deferred drop for render passes to avoid dropping before frame is over

jonatino:
This updates render passes to be dropped in a similar manner as textures. When there are no more references to a render pass (other than itself), we schedule it for deletion that way we're not deleting textures before we use them. This way we guarantee that every render pass/render target will live until the end of the current frame. 

Fixes issue #1002


profan:
I think this looks reasonable to me and it way more closely matches the behaviour people probably expect, though I obviously can't merge it, that's still a @not-fl3 job :eyes:

not-fl3:
Sure sure @profan :D 

Thanks for PR!

PR #1017, opened 2026-02-03
Fix world_to_screen to handle viewport correctly

ESBL-Erik:
## Problem

`screen_to_world` correctly accounts for viewport position and size, but `world_to_screen` always maps to full screen coordinates, ignoring any viewport setting. This makes the two methods non-symmetric when a viewport is set.

## Solution
Added the same viewport handling to `world_to_screen` that `screen_to_world` already has, making them proper inverses of each other.

## Testing
Tested with a Camera2D using a viewport with margins (not at origin). Labels positioned using `world_to_screen` now correctly appear within the viewport region and track world objects properly during pan/zoom.


not-fl3:
Thanks for PR!

PR #1022, opened 2026-03-07
Fix Camera3D aspect when using render target

ingobeans:
Even when rendering to a render target, the aspect that is used for the Camera3D's matrix is based on the screen width and height. This leads to the rendered image being incorrectly stretched and warped when the aspect ratio of the render target doesn't match the screen's.

This PR makes the aspect be taken from the render target if one is used for the camera, else follow the screen size.


not-fl3:
Thanks for PR!

Issue #695, opened 2024-02-01
Color::from_hex(u32) -> Color ignores alpha


asmodehn:
I stumbled upon this recently, lost in confusion coming from debugging colors when alpha was not being what was expected...
https://docs.rs/macroquad/latest/src/macroquad/color.rs.html#95-99

I am guessing alpha should not be ignored here, unless there is a reason to do so (in which case the function should somehow make it explicit) ?

not-fl3:
Fair point, feel free to make a PR! 

Tho I am not quite sure how it should work, 0x3CA7D5 should be a light blue, but with alpha being 0x00 it would a little harder to use from_hex for opaque objects? I do not know really

asmodehn:
I thought a bit about this, and to avoid update problems, I'd like to:
- implement `std::convert::From<u32> for Color` (with alpha)
- modify from_hex to make obvious alpha is always 1... Not sure how though.

Anyway `0x3CA7D5` is 24 bits, so not u32. Not sure how that works...
I ll try to take some time in the following weeks to explore what type conversion is possible safely in rust...

asmodehn:
Looking as SDL, there is a whole structure dedicated to this kind of conversions: https://wiki.libsdl.org/SDL2/SDL_PixelFormat

Not sure what to make of it in our case here... I guess `from_hex` is just too generic/confusing for me...

PR #1024, opened 2026-03-12
Clarify Color::from_hex didn't contains alpha

YuevUwU:
As #695 said, from_hex may be misunderstood as ARGB due to `u32`.
(You can easily see the color to know that it's not RGBA. However, it might not be noticeable if a higher alpha color is used.)

We shouldn't only write `assert_eq!(light_blue.a, 1.00);` to tell devs, it may be outside the IDE's document pane.
We can postpone discussing whether `from_hex` should include alpha.
But hopefully this will alleviate the situation.

Of course, rename or seperate the function name and deprecate it to clarify is better.
But there may have many consideration to do for backward-compatibility.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -111,7 +111,7 @@ impl Color {
         )
     }
 
-    /// Build a color from a hexadecimal u32.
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #1025, opened 2026-03-13
README: fix mentions of the cargo build cache

mochou-p:
as per the [cargo reference](https://doc.rust-lang.org/cargo/reference/build-cache.html "Build Cache - The Cargo Book"), the path of binaries with specific targets, is always in this format: `target/<triple>/<profile>/`

image

the current README incorrectly puts the `<profile>` before the `<triple>`, this PR fixes that

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ rustup target add wasm32-unknown-unknown
 cargo build --target wasm32-unknown-unknown
 ```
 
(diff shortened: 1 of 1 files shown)


cggjaicf (approved):

not-fl3:
Thanks for PR!

PR #1026, opened 2026-03-14
Linux: fix enter/leave mouse state handling

ParriCode:
Mouse enter/leave window events were not handled.
This caused incorrect mouse state when the pointer left the window.

When the mouse left the window, button state was not updated because events outside the window are not received.

The event handling was previously added in the miniquad repository. Here it is integrated into the context behaviuor.

Tested on Linux/X11


not-fl3:
Thanks for PR!

PR #1032, opened 2026-03-31
Don't allocate a `Vec` per pixel in `Font::cache_glyph` (Rust >= 1.53.0)

yuryshulaev:
It took milliseconds on very large glyphs.
Rust 1.53.0 was [released](https://blog.rust-lang.org/2021/06/17/Rust-1.53.0/) on 2021-06-17.
I didn't find any info on MSRV. Is this OK?

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -110,7 +110,7 @@ impl Font {
             Image {
                 bytes: bitmap
                     .iter()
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #1035, opened 2026-04-22
Font Kerning

JMooibroek:
Some fonts contain custom kerning values for two specific characters. Fontdue has a built-in function to retrieve this, but it wasn't implemented yet. I did some simple tests on fonts, and it seems to work without problems.


not-fl3:
Thanks for PR!

PR #1038, opened 2026-05-23
expanded 'clear_input_queue()' to actually clear all relevant input queues

Ludwig-000:
I believe 'keys_pressed' and 'mouse_pressed' are missing from the function, even though they are relevant input queues.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -179,6 +179,8 @@ pub fn clear_input_queue() {
     let context = get_context();
     context.chars_pressed_queue.clear();
     context.chars_pressed_ui_queue.clear();
+    context.mouse_pressed.clear();
+    context.keys_pressed.clear();
 }
 
 /// Detect if the button is being pressed


not-fl3:
What bug does it solves? Not saying that we should not add this, but `clear_input_queue` was a very specific companion function for `get_char_pressed`.

Ludwig-000:
Oh.
It may have just been my interpretation of the function,
but i figured it should clear all "pressed" input events.
Especially since there is, ( to my knowledge ), no other way to clear stuff like mouse pressed. Other than calling 'next_frame()'

not-fl3:
Can you give a usecase when `clear_input_queue()` is needed for normal input?

Ludwig-000:
Oh absolutely.
i am currently working on a little game for a showcase of the library i am working on ( which uses macroquad ).
i'll just give a  simplified version of the code.


```Rust
/// mock code.
/// 
/// 
use macroquad::prelude::*;

async fn main(){
    loop {
        if request_menue(){
            menue().await;
        }
        process_inputs(); // this could, for example, poll the left mouse button for having the character attack.
                          // since LMB was pressed to exit the menue screen, it would still be detected here.
        game_logic();
        next_frame().await;
    }
}

async fn menue(){
    loop {
        // any mouse click would happen on the menue screen
        if is_mouse_button_pressed(MouseButton::Left) && mouse_button_on_quit_button(){
            // but since we want the next frame to be a fresh game frame, we cannot call 'next_frame.await' to reset inputs.
            // while we could use some manual flag to supress inputs, that would be akward.
            return
        }
        do_stuff();
        next_frame().await;
    }
}
```

Ludwig-000:
Alternatively,
the real usecase was right here on line 804
https://github.com/Ludwig-000/pyroquad/blob/main/ExampleGame/TwoDPlarformer.py

though i doubt that's very useful.

not-fl3:
I see, thank you for clarification! :+1:

PR #1043, opened 2026-05-31
Only call glFlush/glFinish if telemetry is enabled (fixes #275)

### LINKED ISSUE #275: High (100%) CPU usage, even with vsync enabled
Seems to be the same issue as #257, but leaving here for tracking purposes in case somebody searches.

On my system, any Macroquad program causes CPU to run at 100%+. This is unique to Macroquad - all the other engines I've tried don't have this problem (Amethist/Tetra/Ggez).

I'm on Ubuntu 20.04, and vsync is enabled.

jmmut:
See #275 

I also have linux with nvidia and for all macroquad programs I get ~96% CPU unless I put sleeps (which shouldn't be done inside async).

After the research done by the people in the issue and me, it seems to me that glFinish is not really needed, so this PR only keeps it if telemetry is enabled. I also made a getter for `telemetry::is_enabled()`.

Here is a CPU usage comparison of the "basic_shapes" example:

- current master ed9cdbd
mq_glfinish_ed9cdbd_before

---

- this PR dce417a
mq_glfinish_dce417a_after

I also tried this in wasm for my game and it seems to work with lowish CPU, but I didn't make a comparison so I don't know if before it was the same or higher.

---

Let me know if you want different format or different comments. Also it's ok if you don't want to merge, I did my own fork to try out other stuff.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -802,10 +802,17 @@ impl EventHandler for Stage {
             get_context().frame_time = date::now() - get_context().last_frame_time;
             get_context().last_frame_time = date::now();
 
+            // glFinish waits until the drawing is done. See https://registry.khronos.org/OpenGL-Refpages/gl4/html/glFinish.xhtml.
+            // Some drivers do this by a busy loop which increases CPU usage to close to 100%.
+            // For discussion see https://github.com/not-fl3/macroquad/issues/275.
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks for PR!

PR #1059, opened 2026-08-04
Remove `color_u8!` from the docs

Vellumic:
Initially I considered adding `#[deprecated]`, but found out that `color_u8!` can handle mixed `u8` and `f32`, whereas `Color::from_rgba` strictly requires `u8` meaning they aren't 100% interchangeable. Forcing a deprecation warning or trying to cast types `as u8` in existing code would introduce subtle rounding bugs and break tests.

Using `#[doc(hidden)]` is the best middle ground here. It improves DX by hiding the legacy macro from `docs.rs` so new users won't stumble upon it, while preserving 100% backward compatibility for existing code.


not-fl3:
Thanks for PR!

PR #1063, opened 2026-08-16
text: avoid spurious font atlas growth

### LINKED ISSUE #1054: Font atlas doubles spuriously at a 1px row-boundary condition; random-order repacking makes growth nondeterministic (up to 32768x32768 + overflow panics)
## Summary

`Atlas::cache_sprite` (src/text/atlas.rs) has a 1px boundary condition at the end of a row that doubles the atlas texture even when there is plenty of free space, instead of simply breaking to the next line. Combined with the fact that regrowing repacks sprites in `HashMap` iteration order (i.e. random order), atlas growth becomes nondeterministic and can cascade up to 32768x32768 for glyph sets whose actual area fits comfortably in 2048x2048.

Present in 0.4.15 and current master:
https://github.com/not-fl3/macroquad/blob/master/src/text/atlas.rs

## Bug 1: spurious grow at the row boundary

```rust
let x = if self.cursor_x + (width as u16) < self.image.width {
    // stay in current line, place at x = cursor_x + GAP
    ...
} else {
    // line break
    ...
};
let y = self.cursor_y;

// texture bounds exceeded
if y + sprite.height > self.image.height || x + sprite.width > self.image.width {
    // -> drain sprites, double the atlas, recache everything
```

The "fits in current line" check ignores the `GAP` margin, but the actual placement is `x = cursor_x + GAP`. When `cursor_x == image.width - width - 1`, the first check passes, yet `x + width > image.width` — so instead of a line break, the code falls into the "texture bounds exceeded" branch and **doubles the whole atlas**.

For random glyph widths this fires with probability ~`1 / (avg_glyph_width + 2*GAP)` per completed row, so any workload that fills more than a handful of rows (e.g. caching CJK glyphs at several font sizes) hits it sooner or later.

## Bug 2: random-order repacking makes it cascade

On grow, sprites are drained from the `HashMap` and re-cached in iteration order, i.e. effectively shuffled. Mixed glyph heights in one row waste vertical space, and — more importantly — every repack fills many rows again, each of which re-rolls the Bug 1 dice. One spurious grow therefore tends to trigger further ones, recursively, with a different outcome on every run.

## Observed behavior

Instrumenting `cache_sprite` with an `eprintln!` on the grow path and caching ~340 distinct glyphs (Japanese UI text, NotoSansJP) at 16 font sizes via `measure_text`:

```
run 1: 512->1024, 1024->2048, 2048->4096                                (final 4096)
run 2: 512->1024, 1024->2048, 2048->4096, 4096->8192, 8192->16384      (final 16384)
run 3: 512->1024, 1024->2048, 2048->4096                                (final 4096)
```

Same input every run. Every grow beyond 1024 triggers at `x ≈ image.width` with a small sprite (e.g. `trigger=9x10 x=4088` on a 4096-wide atlas), i.e. none of them is a real out-of-space condition — the whole glyph set fits in 2048x2048.

Consequences when the cascade goes far enough:

* **Native, debug build**: at 32768x32768, miniquad's `TextureFormat::size` computes `4 * width * height` in `u32`, which is exactly 2^32 → multiply-with-overflow panic at startup.
* **wasm32**: `usize` is 32-bit, so `Image::gen_image_color`'s `width * height * 4` overflows to 0 → zero-length buffer → out-of-bounds panic → frozen black canvas.
* Even when it doesn't overflow, a 16384x16384 RGBA atlas costs 1 GB of RAM + VRAM for a few thousand small glyphs.
* Each unnecessary grow also deletes/recreates the atlas texture mid-frame, which corrupts the current text batch on GL — this may well be a contributing cause of the "black rects" reports in #1006.

## Fix (tested)

Two small changes in `cache_sprite` fix both problems; with them the same workload packs deterministically into 2048x2048 on every run:

```rust
// 1. account for the gaps in the "fits in current line" check,
//    so the right edge can never cross the texture border:
let x = if self.cursor_x as u32 + width as u32 + Self::GAP as u32 * 2
    <= self.image.width as u32
{
    ...

// 2. repack tallest-first instead of HashMap order, making growth
//    deterministic and near-minimal:
let mut sprites = self.sprites.drain().collect::<Vec<_>>();
sprites.sort_by_key(|(_, sprite)| {
    std::cmp::Reverse((sprite.rect.h as u32, sprite.rect.w as u32))
});
```

## Reproduction

```rust
use macroquad::prelude::*;

#[macroquad::main("atlas_grow")]
async fn main() {
    // any font with CJK coverage, e.g. NotoSansJP-Regular.ttf
    let font = load_ttf_font("NotoSansJP-Regular.ttf").await.unwrap();
    let glyphs: String = ('あ'..='ん').chain('一'..='下').collect(); // a few hundred chars
    for size in [11u16, 12, 13, 14, 16, 17, 18, 19, 20, 22, 24, 25, 29, 31, 34, 38] {
        measure_text(&glyphs, Some(&font), size, 1.0);
    }
    // add an eprintln! to the grow branch of Atlas::cache_sprite (or watch RSS):
    // final atlas size differs from run to run; with the fix it is always 2048.
    loop { next_frame().await }
}
```

Happy to open a PR with the fix if that helps.

Environment: macroquad 0.4.15 (also inspected current master), Windows 11, OpenGL backend — though both bugs are pure packing logic and platform-independent.

atirna:
## Summary

`Atlas::cache_sprite` places a sprite at `cursor_x + GAP`, but its row-fit check did not reserve that margin. A sprite that crossed the right edge by one pixel was therefore placed out of bounds and grew the atlas instead of starting a new row.

This checks the placement bound in `u32`, keeps exact right-edge placement valid, and repacks cached sprites in a deterministic order when the atlas grows.

Fixes #1054

## Verification

- `cargo test --lib`
- `cargo build --all-targets`
- `cargo check --lib --target wasm32-unknown-unknown`

### DIFF
--- a/src/text/atlas.rs
+++ b/src/text/atlas.rs
@@ -125,7 +125,8 @@ impl Atlas {
     pub fn cache_sprite(&mut self, key: SpriteKey, sprite: Image) {
         let (width, height) = (sprite.width as usize, sprite.height as usize);
 
-        let x = if self.cursor_x + (width as u16) < self.image.width {
+        let x = if self.cursor_x as u32 + width as u32 + Self::GAP as u32 <= self.image.width as u32
+        {
             if height as u16 > self.max_line_height {
                 self.max_line_height = height as u16;
             }
@@ -143,7 +144,6 @@ impl Atlas {
         // texture bounds exceeded
         if y + sprite.height > self.image.height || x + sprite.width > self.image.width {
             // reset glyph cache state
-            let sprites = self.sprites.drain().collect::<Vec<_>>();
             self.cursor_x = 0;
             self.cursor_y = 0;
             self.max_line_height = 0;
@@ -161,6 +161,25 @@ impl Atlas {
                 Image::gen_image_color(new_width, new_height, Color::new(0.0, 0.0, 0.0, 0.0));
 
             // recache all previously cached symbols
+            // sprites are repacked tallest-first, ties broken by the previous
+            // rect position: repacking in `HashMap` iteration order is
+            // effectively random, which both wastes space on rows of mixed
+            // heights and makes the atlas growth nondeterministic from run to
+            // run. Ties must be broken deterministically as well — rect
+            // positions are unique and themselves a product of the previous
+            // deterministic packing, so this is a total order and the
+            // key-to-rect mapping stays identical for every atlas instance.
+            let mut sprites = self.sprites.drain().collect::<Vec<_>>();
+            sprites.sort_by(|(_, a), (_, b)| {
+                let key = |sprite: &Sprite| {
+                    (
+                        std::cmp::Reverse((sprite.rect.h as u32, sprite.rect.w as u32)),
+                        sprite.rect.y as u32,
+                        sprite.rect.x as u32,
+                    )
+                };
+                key(a).cmp(&key(b))
+            });
             for (key, sprite) in sprites {
                 let image = old_image.sub_image(sprite.rect);
                 self.cache_sprite(key, image);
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

